PR #4, opened 2015-01-08
Add filter for using site names in directory structure.

### LINKED ISSUE #3: [Request] Filter for using site name in directory structure
As of version 3.6.0, for a specific site of a multi-site installation, the PDF templates reside in `/%uploads%/PDF_EXTENDED_TEMPLATES/%site_name%/`.  
It would be great to be able to disable this easily, and make the templates live in `/%uploads%/PDF_EXTENDED_TEMPLATES/` (just like it was before, inside the theme).

A real-life scenario where this filter is needed:  
Suppose you are working together with a number of developers on a project using some sort of version control. Each developer is having an individual local testing environment. Then there is a testing and a productive server. As the site name is part of the directory structure (by default), sooner or later, the project repository will have multiple (redundant) entries, like so:
* `/%uploads%/PDF_EXTENDED_TEMPLATES/localhost_project_x/*`
* `/%uploads%/PDF_EXTENDED_TEMPLATES/project_x_dev/*`
* `/%uploads%/PDF_EXTENDED_TEMPLATES/project_x_testsystem_com/*`
* `/%uploads%/PDF_EXTENDED_TEMPLATES/project_x_com/*`

In all of these folders, you find the same templates.

If one were able to exclude the site name (meaning that all sites use the exact same templates), there would only be a single folder `/%uploads%/PDF_EXTENDED_TEMPLATES/` containing these files.

I will commit a simple fix (as well as the according pull request) containing a filter `gfpdf_use_site_name_in_directory_structure` to decide if one would like to use the site names or not. By default, I will leave the functionality as it is now. And not using the site names in the directory structure is as simple as this:

    add_filter( 'gfpdf_use_site_name_in_directory_structure', '__return_false' );

tfrommen:
Fix #3.

### DIFF
--- a/helper/data.php
+++ b/helper/data.php
@@ -52,17 +52,23 @@ class GFPDFE_DATA
     public function set_directory_structure()
     {
         $upload_dir = PDF_Common::get_upload_dir();
-        $site_name  = PDF_Common::get_site_name();
+
+        if ( apply_filters( 'gfpdf_use_site_name_in_directory_structure', TRUE ) ) {
+            $site_name = PDF_Common::get_site_name();
+            $site_name = trailingslashit( $site_name );
+        } else {
+            $site_name = '';
+        }
 
         $this->template_location          = $upload_dir['basedir'] . '/' . PDF_SAVE_FOLDER . '/';
-        $this->template_site_location     = $upload_dir['basedir'] . '/' . PDF_SAVE_FOLDER . '/' . $site_name . '/';
-        $this->template_save_location     = $upload_dir['basedir'] . '/' . PDF_SAVE_FOLDER . '/' . $site_name . '/output/';
-        $this->template_font_location     = $upload_dir['basedir'] . '/' . PDF_SAVE_FOLDER . '/' . $site_name . '/fonts/';
-        
+        $this->template_site_location     = $upload_dir['basedir'] . '/' . PDF_SAVE_FOLDER . '/' . $site_name;
+        $this->template_save_location     = $upload_dir['basedir'] . '/' . PDF_SAVE_FOLDER . '/' . $site_name . 'output/';
+        $this->template_font_location     = $upload_dir['basedir'] . '/' . PDF_SAVE_FOLDER . '/' . $site_name . 'fonts/';
+
         $this->template_location_url      = $upload_dir['baseurl'] . '/' . PDF_SAVE_FOLDER . '/';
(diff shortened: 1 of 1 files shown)


blueliquiddesigns:
Thanks for the pull request. 

It looks like a good idea. When I have time next week I'll review the changes to ensure there are no conflicts and merge it into the next release.

blueliquiddesigns:
We've had a few other users have similar issues to yours.

Currently we are thinking this should become the default behaviour for non-multisite installations (templates and config in the root of the PDF_EXTENDED_TEMPLATES folder), while multisites should use the site ID for the folder name rather than the domain name.

Baring any unforeseen problems this should fix the issue between different folder names for staging and production in both single-site and multi-site WordPress environments. It's just a shame we didn't have feedback like this while running our beta program.

tfrommen:
Hi there!

Using the site ID instead of the domain name should do the trick, that's right. But wouldn't it be better to just use the built-in function `wp_upload_dir` instead of your `PDF_Common::get_upload_dir` function? The main part of it is a mere copy of `wp_upload_dir` anyway, isn't it?

The templates for a _secondary_ site with ID 2 will then reside in `wp-content/uploads/sites/2/PDF_EXTENDED_TEMPLATES` instead of `wp-content/uploads/PDF_EXTENDED_TEMPLATES/2`. The templates of the main site (no matter if using single-site or multi-site) will always stay in `wp-content/uploads/PDF_EXTENDED_TEMPLATES`.

blueliquiddesigns:
Thanks for the feedback.

One of the reasons (among many) for moving the configuration out of the theme folder was to consolidate the files for multisite users. It makes it easier from a management standpoint to apply templates across multiple websites that share identical forms (something a lot of our users do).

It also makes sense from a version control point of view, too. Most users using a staging/production environment will have the /uploads/ directory in their .gitignore file. It's much simpler to override this behaviour if you only have to add a single folder once. 

Also, with such a big change to the folder structure in the 3.6 release we don't want to then move a number of folders again and have our user-base more confused. The change of the site directory names to numbers will be bad enough, but the benefits in this case out way the downsides.

blueliquiddesigns:
[Issue fixed in 3.7 Beta](https://developer.gravitypdf.com/news/gravity-pdf-3-7-0-beta-1-released/). Give it a test and let me know if there are any issues.

tfrommen:
Seems to be working just fine.
Thank you for the fix, and the post. ;)

Issue #38, opened 2015-11-02
Convert PDF Template Doctype to HTML5


blueliquiddesigns:
Replace `<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
  "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">` doctype with `<!DOCTYPE html>`.

Issue #57, opened 2015-11-16
Handle Core Header, Footer and Background independant of PDF template


blueliquiddesigns:
Removes a step in the user custom template development to including header/footer/backgrounds

Issue #42, opened 2015-11-04
Pre-transform Header and Footer settings with `$gfpdf->misc->fix_header_footer()`


blueliquiddesigns:
Before the core header/footer settings hit the template we should pre-transform them so custom template users don't have to.

Issue #41, opened 2015-11-04
Pass `$gfpdf` directly to PDF Templates


blueliquiddesigns:
Prevent the need for `global $gfpdf` and pass in the variable straight to templates.

PR #59, opened 2015-11-16
Steamline Template Variables and Core Styles

### LINKED ISSUE #39: Automatically Set font styles based on PDF Settings
At the moment font styles are manually added to PDF templates like custom PDF settings. To make it easier when custom templating let's automatically set up these before the PDF is loaded.

blueliquiddesigns:
- Steamline the PDF templates so core styles like font, colour, size and headers/footers don't need to be manually set by the user
- Make $gfpdf available to templates automatically
- Preprocess header/footer settings so users don't need to do so via the template

Fixes #39, #41, #57, #42

Remove the "!important" tag from the page font color

### DIFF
--- a/initialisation/templates/zadani.php
+++ b/initialisation/templates/zadani.php
@@ -23,6 +23,7 @@ if ( ! class_exists( 'GFForms' ) ) {
  * $lead (alias of $entry)
  * $form_data (The processed entry data stored in an array)
  * $settings (the current PDF configuration)
+ * $gfpdf (the main Gravity PDF object containing all our helper classes)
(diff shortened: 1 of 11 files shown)


blueliquiddesigns:
Before this can be merged unit tests need to be written for the new functionality

blueliquiddesigns:
Also fixes #38

blueliquiddesigns:
Update documentation to reflect new changes in PR. See https://github.com/GravityPDF/v4-documentation/commit/d7c5428b262de262cb0da416c7135e7a0b4a9e7d

Issue #370, opened 2016-05-11
Enable config setting for disabling CSS Ready Class Support in PDF


jakejackson1:
Leave CSS Ready Class Support on by default in PDFs but give users the option to turn this off.

PR #376, opened 2016-05-12
Add switch to disable CSS Ready Class support

jakejackson1:
This allows developers more freedom to style the PDF as they see fit.
Ensure PDF templates have CSS Ready Class support enabled by default


jakejackson1:
Resolves #370

PR #387, opened 2016-05-24
Apply HTML escaping during value() method

### LINKED ISSUE #385: Check v3 vs v4 escape handling in $form_data array
Most importantly, check the following text in text and paragraph fields:

`<test> "this" 'stuff' & Other Stuff`

jakejackson1:
Instead of applying esc_html() during the html() method, we're going to apply it at the value() method so the $form_data array is appropriately protected as well.
This does break backwards compatibility and make it more difficult for users to apply conditional logic in custom templates (as there's encoded characters to consider - < > ' " &) but we'll note this in the developer documentation and upgrade guide.

Resolves #385

### DIFF
--- a/src/helper/fields/Field_Address.php
+++ b/src/helper/fields/Field_Address.php
@@ -121,14 +121,10 @@ class Field_Address extends Helper_Abstract_Fields {
 			$address[] = $data['country'];
 		}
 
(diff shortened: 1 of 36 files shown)


jakejackson1:
Addition appropriate documentation about issue.

PR #468, opened 2016-07-29
Set master password to null so it's automatically generated by mpdf

### LINKED ISSUE #467: PDF Security Master Password Issue
Mpdf automatically generates a master password when passing `null` as the third parameter in `SetProtection()`. It does not work if you pass an empty string (which is what we are doing now).

jakejackson1:
Fixes #467


jakejackson1:
Client confirmed this fix works on Mac.

Issue #486, opened 2016-08-23
Create an advanced template manager


jakejackson1:
Utlise the design of the theme manager, an create a JS-powered template manager to view, select, delete or install new templates. 

This will compliment the new template shop which will be timed with this release.

odyright:
Hello, what about the final release?

jakejackson1:
Hi @odyright, this feature is currently scheduled for Gravity PDF 4.1 and will be released at that time (there is no time line for this).

odyright:
ok. Now, Is Gravity pdf support html5 & Css3 or such new design like Maaterial design? i asked, because i have a custum template made with it.

jakejackson1:
> Now, Is Gravity pdf support html5 & Css3 or such new design like Maaterial design? i asked, because i have a custum template made with it. 

There's [very limited support for HTML5 / CSS3 features](http://mpdf.github.io/css-stylesheets/supported-css.html). However, this is not the correct location to ask such as question. [Please use our support ticket system](https://gravitypdf.com/support/). GitHub issues are for reporting bugs related to Gravity PDF.

odyright:
ok thanks!

PR #505, opened 2016-10-21
Create a advanced PDF template selected using ReactJS

jakejackson1:
The design is based on the backbone-powered Theme selector, but is powered by
ReactJS and JSX, instead of Backbone and Underscore.

It allows the end user to search, view and select the specific PDF template they
want to use. This is just stage 1. Stage 2 will allow the user to install and
delete PDF templates too.

We are using Webpack and Babel to write modern JS and transpile it down into
ES5. All new JS is tested with Mocha, Chia and Sinon and gets run using Karma.

Along with the JS changes, all template code was refracted into Helper_Templates.
The code being moved isn't backwards compatible with v4. Appropriate unit tests
were also written.

Travis has been updated to run our JS tests on the Multisite instance.

This PR is the View layer for https://github.com/GravityPDF/gravity-pdf/issues/486. The install and delete functionality still needs to be done.


jakejackson1:
Need to add GPL / copyright notices to all JS files. Should also add JS-equivalent to docblocks.

PR #536, opened 2016-11-22
Move our Abstract classes from /helper/abstract/ to /abstraction/

### LINKED ISSUE #423: Move the "abstract", "fields" and "interface" directories out of "helper"
These directories stand on their own and are important to the operation of the software. They should sit in the root `src` directory.

jakejackson1:
We dropped `Helper_` from the file name and changed the namespace to `GFPDF/Abstraction/`.

Partially resolves #423


jakejackson1:
Not moving forward with this. Refer to https://github.com/GravityPDF/gravity-pdf/issues/423.

PR #571, opened 2017-01-17
Resolve mime-issues with WordPress 4.7.1 by disabling check for TTF files

### LINKED ISSUE #570: Cannot upload font to Media Library after WordPress 4.7.1 Update
See https://core.trac.wordpress.org/ticket/39550 for details about this issue

jakejackson1:
Fixes #570


jakejackson1:
WordPress 4.7.3 should fix the problem if we change the  TTF mime type to "application/x-font-ttf". Since it uses `finfo_open` with PHP's built-in magic file (maybe they should ship their own?), it should be okay - potentially.

jakejackson1:
This patch ensures compatibility with changes to WordPress 4.7.3 (see https://core.trac.wordpress.org/ticket/39550)

PR #780, opened 2018-03-12
src/view/View_PDF.php

MenashayAu:
/* Try and display our HTML */
       /* Menashay Givoni (http://customiseit.com.au) 18-March-2018
        * Added support for repeator add-on 
       */
        try {
       	   
       	    /*
       	    * Menashay Givoni (http://customiseit.com.au) 18-March-2018
       	    * Process repeater field separately
       	    */ 	
            if($field -> type === 'repeater') {
	        echo($field -> get_value_export($entry, '', 'pdf', false));
            } else {
	    	
            /* Only load our HTML if the field is NOT empty, or the $empty config option is true */
            if ( ! $class->is_empty() || $show_empty_fields === true ) {
                /* Load our legacy CSS class names */
                if ( $load_legacy_css === true ) {
                    $this->load_legacy_css( $field );
                }

                /**
                 * Add CSS Ready Class Float Support to mPDF
                 * Open a HTML container if needed
                 */
                $container->generate( $field );

                echo ( $field->type !== 'section' ) ? $class->html() : $class->html( $show_section_description );
            } else {
                /* To prevent display issues we will output the column markup needed */
                $container->maybe_display_faux_column( $field );
            }
        }
        } catch ( Exception $e ) {
            $this->log->addError( 'PDF Generation Error', [
                'field'     => $field,
                'entry'     => $entry,
                'config'    => $config,
                'form_id'   => $form['id'],
                'exception' => $e->getMessage(),
            ] );
        }
    }

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -85,6 +85,12 @@ Also, if you enjoy using the software [we'd love it if you could give us a revie
 
 == Changelog ==
 
+= 4.3.1 =
+* Bug: Restrict Gravity PDF JavaScript to the correct PDF pages (GH#693)
+* Bug: Fix PHP5.2 activation error (GH#697)
+* Bug: Fix RTL issue with Chosen Select library (GH#698)
+* Bug: Fix PDF Product table pricing issue by using the pre-calculated price field for the unit price (GH#699)
+
 = 4.3.0 =
 * Feature: Add support for Gravity Perks E-Commerce Add-on (GH#671)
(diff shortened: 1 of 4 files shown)


jakejackson1:
Sorry. This PR [doesn't meet the requirements outlined in our CONTRIBUTING.md guide](https://github.com/GravityPDF/gravity-pdf/blob/development/CONTRIBUTING.md). Feel free to submit a new PR that does meet these guidelines.

PR #791, opened 2018-05-15
Add native support for the Slim Image Cropper for Gravity Forms plugin

jakejackson1:
Will display uploaded images in the PDF. Recommended users set a maximum width / height to 1200,1200 or less to ensure no issues with PDF generation due to large file size.


jakejackson1:
Before merging, need to check the Slim Post field as well.

jakejackson1:
Should also run the path/URL through `esc_url`

Issue #842, opened 2018-11-07
Add automatic deployment from Travis to .org when tagging releases


jakejackson1:
If I never have to manually do an SVN push again I'll be extremely happy. 

Here's some starting points:

https://deliciousbrains.com/deploying-wordpress-plugins-travis/
https://kuzmi.ch/articles/wordpress-plugin-and-theme-development-using-git-only/

PR #845, opened 2018-11-15
Add automatic WP deployment on new tags

jakejackson1:


jakejackson1:
Refine so the new tag is created remotely, then checked out, files added / deployed. This will save the entire SVN directory getting checked out.

See https://stackoverflow.com/a/1155233/1614565

jakejackson1:
Resolves https://github.com/GravityPDF/gravity-pdf/issues/842

PR #883, opened 2019-02-27
881 Add Code Sniff Support

saggonzales:


jakejackson1:
Continued in https://github.com/GravityPDF/gravity-pdf/pull/887

Issue #728, opened 2017-10-31
Refactor ReactJS Code to remove Immutable


jakejackson1:
Unminified, this extension adds 139KB and makes up 9.79% of the bundle size. 

Can make use of ES6 spread operator to easily create/update immutable objects

PR #888, opened 2019-03-04
Refactor ReactJS Code to remove Immutable

debbieellis7:
This is a work in progress.

https://github.com/GravityPDF/gravity-pdf/issues/728


debbieellis7:
Moved into R2 #893

PR #893, opened 2019-03-07
Refactor ReactJS Code to remove Immutable [R2]

debbieellis7:


src/assets/js/react/components/Template/TemplateActivateButton.js
```diff
@@ -44,7 +44,6 @@ export class TemplateActivateButton extends React.Component {
    * @since 4.1
    */
   static propTypes = {
-    template: PropTypes.object,
```
jakejackson1: The `template` prop is still being passed to the component, so wouldn't this propType still apply?

src/assets/js/react/components/Template/TemplateActivateButton.js
```diff
@@ -79,7 +80,7 @@ export class TemplateActivateButton extends React.Component {
   }
 }
 
-/**
+/**TemplateActivateButton
```
jakejackson1: `/**` should be on a line all by itself

debbieellis7: Fixed

Issue #889, opened 2019-03-05
Code Splitting / Caching / Lazy Loading JavaScript


jakejackson1:
To reduce the size of the JS being loaded per page, we should look at both code splitting and lazy loading the React JS.

Let's start with the easy stuff like using multiple entry points to split the `/node_modules/` imported files into a `vendors.js` file, then split modules up by their page(s) - e.g the Help module. 

Once we've got that dialled, we can look at adding cache-busting hashes. 

Finally, we can play around with lazy loading JavaScript. 

While you are working on this, you may also want to explore a cleaner Webpack config file, and separate the development and production files. To give you some ideas, this setup wasn't bad: https://github.com/itzikbenh/WordPackFour

Note: You'll likely open a PR for each type. I've just included this all as a single issue as they are closely related to Webpack.

Additional reading:
https://webpack.js.org/guides/code-splitting/
https://webpack.js.org/guides/lazy-loading/
https://hackernoon.com/the-100-correct-way-to-split-your-chunks-with-webpack-f8a9df5b7758
https://reacttraining.com/react-router/web/guides/code-splitting
https://webpack.js.org/guides/public-path/#on-the-fly (<-- the on the fly `publicPath` assignment could 
be the solution we're looking for to allow lazy loading). 
https://webpack.js.org/guides/caching/ (could be viable in WordPress via https://danielshaw.co.nz/wordpress-cache-busting-json-hash-map/ and injecting the correct path via `wp_localize_script()`)

PR #938, opened 2019-05-17
React lazy loading

debbieellis7:
## Description
Webpack, Babel, and React setup for lazy loading.
Issue: #889
This modification is part of issue #889 

## Checklist:
- [x] I've tested the code.


webpack.config.js
```diff
@@ -2,6 +2,8 @@ const webpack = require('webpack')
 const path = require('path')
 const UglifyJSPlugin = require('uglifyjs-webpack-plugin')
 const PROD = (process.env.NODE_ENV === 'production')
+const pluginDirectory = path.basename(__dirname)
+const chunkPath = `/wp-content/plugins/${pluginDirectory}/dist/assets/js/`
```
jakejackson1: We need to be able to get rid of `/wp-content/plugins/` in the config.

debbieellis7: I think that's the simplest way possible. Here is another way.

const pluginDirectory = path.basename(__dirname)
const wpPath = path.basename(path.resolve('../../'))
const plgPath = path.basename(path.resolve('../'))
const publicPath = `/${wpPath}/${plgPath}/${pluginDirectory}/dist/assets/js/`

jakejackson1:
@debbieellis7 I've made some modifications to this PR to allow the public path to be defined at runtime. This resolves the issue I brought up in my previous review. 

Could you run through it, fix up any linting issues, verify it works as expected, and squash ect?

debbieellis7:
> @debbieellis7 I've made some modifications to this PR to allow the public path to be defined at runtime. This resolves the issue I brought up in my previous review.
> 
> Could you run through it, fix up any linting issues, verify it works as expected, and squash ect?

All good!

Issue #940, opened 2019-06-02
Add JS-linting to package, as well as Travis. 


jakejackson1:
We want to have JS-based linting that can be run on the package automatically when TravisCI runs.

jakejackson1:
These should help:

https://medium.com/@RossWhitehouse/setting-up-eslint-in-react-c20015ef35f7
https://gist.github.com/ronapelbaum/ef3c3459f01b002d0d719ca205401174

PR #947, opened 2019-06-10
Add JS-linting to package & Travis

debbieellis7:
## Description
Issue: #940 
JS-based linting that can be run on the package automatically when TravisCI runs.

## Checklist:
- [ ] I've tested the code.


.eslintrc
```diff
@@ -5,7 +5,7 @@
     "ecmaFeatures": {
       "impliedStrict": true,
       "jsx": true,
-      "experimentalObjectRestSpread": true,
+      "experimentalObjectRestSpread": true
```
jakejackson1: ecmaFeatures.experimentalObjectRestSpread is deprecated https://eslint.org/docs/user-guide/configuring

.eslintrc
```diff
@@ -23,7 +23,13 @@
     "react"
   ],
   "extends": [
-    "eslint:recommended",
+//    "eslint:recommended",
```
jakejackson1: You should review https://eslint.org/docs/rules/ and see exactly which rules we should apply. `eslint:recommended` is a good starting point though.

PR #989, opened 2019-10-17
Resolves an issue redirecting to the PDF in AJAX mode

jakejackson1:
## Description

When the AJAX redirect was used, GF generated the redirect JS string and passed it to the filter we tapped into. Because it was a string, it was being handled by our text confirmation code. This patch corrects that problem.

## Testing instructions
1. Embed a form (any) on a WordPress Page or Post and enable AJAX mode
1. Ensure a PDF is configured on the form, then copy the [gravitypdf] shortcode show on the PDF List page
1. Open up the default Confirmation, select Redirect mode, then paste in the shortcode and save.
1. Submit the form embedded on your post/page, it'll hang and the browser Console will show a JS error
1. When checking out this branch, the AJAX Confirmation will work correctly. 

## Checklist:
- [x] I've tested the code.

## Additional Comments 
When testing, you should double check the non-AJAX redirect, page redirect, and text redirect also work as expected and the shortcode generates correctly. 

I may need to add an additional unit test to cover this change, but will review after internal testing.

### DIFF
--- a/src/Helper/Helper_Abstract_Pdf_Shortcode.php
+++ b/src/Helper/Helper_Abstract_Pdf_Shortcode.php
@@ -164,7 +164,7 @@ abstract class Helper_Abstract_Pdf_Shortcode extends Helper_Abstract_Model {
 	public function gravitypdf_confirmation( $confirmation, $form, $entry ) {
 
 		/* check if confirmation is text-based */
-		if ( ! is_array( $confirmation ) ) {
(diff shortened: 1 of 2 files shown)


debbieellis7:
Form spinner is stuck spinning after a successful form submission and PDF download.

jakejackson1:
@debbieellis7 thanks. For the moment we won't worry about fixing the AJAX spinner. I'd like to push this fix down the track instead https://github.com/GravityPDF/gravity-pdf/issues/990

Issue #991, opened 2019-10-23
Fix Welcome Screen in WordPress 5.3-RC2


jakejackson1:
![image](https://user-images.githubusercontent.com/2918419/67380239-c09f1100-f5d5-11e9-818a-2796af17774f.png)

The fix for this needs to be backwards compatible right up to WordPress 4.8, which we still support.

PR #996, opened 2019-10-30
Fixed Welcome Screen (backwards compatible right up to WordPress 4.8)

debbieellis7:
## Description

Issue: #991 
This fix the issue above which is backwards compatible right up to WordPress 4.8.

## Checklist:
- [x] I've tested the code.


debbieellis7:
> We should limit the columns to screens wider than 600px. If it's smaller than this, everything displays in a single column.

done

Issue #999, opened 2019-11-03
Check for WP_Error before handling Confirmations. 


jakejackson1:
```
Uncaught Error: Cannot use object of type WP_Error as array in /var/www/vhosts/swiercs.nl/httpdocs/wp-content/plugins/gravity-forms-pdf-extended/src/model/Model_Shortcodes.php:255
Stack trace:
#0 /var/www/vhosts/swiercs.nl/httpdocs/wp-includes/class-wp-hook.php(288): GFPDF\\Model\\Model_Shortcodes->gravitypdf_confirmation('<div id='gform_...', Array, Object(WP_Error))
#1 /var/www/vhosts/swiercs.nl/httpdocs/wp-includes/plugin.php(208): WP_Hook->apply_filters('<div id='gform_...', Array)
#2 /var/www/vhosts/swiercs.nl/httpdocs/wp-content/plugins/gravityforms/gravityforms.php(5458): apply_filters('gform_confirmat...', '<div id='gform_...', Array, Object(WP_Error), false, NULL, NULL, NULL, NULL, NULL, NULL, NULL)
#3 /var/www/vhosts/swiercs.nl/httpdocs/wp-content/plugins/gravityforms/form_display.php(1567): gf_apply_filters('gform_confirmat...', '<div id='gform_...', Array, Object(WP_Error), false)
#4 /var/www/vhosts/swiercs.nl/httpdocs/wp-content/plugins/gf-mollie-by-indigo/class-g...
'
```

jakejackson1:
@ayenmot we need to add guarded checks (IF statements) against [WP_Error](https://developer.wordpress.org/reference/classes/wp_error/) in the following methods:

https://github.com/GravityPDF/gravity-pdf/blob/development/src/Helper/Helper_Abstract_Pdf_Shortcode.php#L167
https://github.com/GravityPDF/gravity-pdf/blob/development/src/Helper/Helper_Abstract_Pdf_Shortcode.php#L186
https://github.com/GravityPDF/gravity-pdf/blob/development/src/Helper/Helper_Abstract_Pdf_Shortcode.php#L338

See https://code.tutsplus.com/tutorials/wordpress-error-handling-with-wp_error-class-i--cms-21120 for details on how WP_Error works

jakejackson1:
Looks like this needs some more work: https://github.com/GravityPDF/gravity-pdf/commit/a3a4739feaf8f76379ff4ec39e560a0aafe5c75a

[...]

PR #1003, opened 2019-11-07
Added a filter to check if the first parameter isn't a WP_Error

ayenmot:
## Description

Added a filter that checks the first parameter passed isn't a WP_Error
#999 

## Checklist:
- [ ] I've tested the code.

### DIFF
--- a/src/Helper/Helper_Abstract_Pdf_Shortcode.php
+++ b/src/Helper/Helper_Abstract_Pdf_Shortcode.php
@@ -163,7 +163,16 @@ abstract class Helper_Abstract_Pdf_Shortcode extends Helper_Abstract_Model {
 	 */
 	public function gravitypdf_confirmation( $confirmation, $form, $entry ) {
 
-		/* check if confirmation is text-based */
+		/**
+		 * Do nothing if WP_Error is passed
+		 *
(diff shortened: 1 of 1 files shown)


src/Helper/Helper_Abstract_Pdf_Shortcode.php
```diff
@@ -30,6 +30,7 @@ abstract class Helper_Abstract_Pdf_Shortcode extends Helper_Abstract_Model {
 
 	/**
 	 * Set this constant to the Shortcode ID you're using
+	 *
```
jakejackson1: Remove this. It's unrelated to this PR.

src/Helper/Helper_Abstract_Pdf_Shortcode.php
```diff
@@ -164,6 +165,10 @@ protected function get_pdf_config( $entry_id, $pdf_id ) {
 	public function gravitypdf_confirmation( $confirmation, $form, $entry ) {
 
 		/* check if confirmation is text-based */
+		if ( is_wp_error( $confirmation ) ) {
```
jakejackson1: This should go above the comment on line 167. There should also be an inline comment explaining why we are checking for a WP Error with a link back to GitHub. This applies to all uses in this file.

PR #1048, opened 2020-02-21
Do not generate field container for skipped types

FPCSJames:
## Description
Previously, the `generate` function of `Helper_Field_Container` was not doing anything with the return value of `process_skipped_fields`. This causes field types which should not have a field container wrapped around them to have one wrapped around them.

This patch checks the return value of `process_skipped_fields` and stops generation if `true`. This will not leave the previous container open, because `process_skipped_fields` closes it if a skipped field is hit.

## Testing instructions
1. Add a field type to the skip_fields array via the `gfpdf_container_skip_fields` filter, enable a display of a skipped field type (such as HTML) via the settings, or create a class for a field type such as page.
2. Observe that the field container HTML is no longer erroneously being generated.

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/Helper/Helper_Field_Container.php
+++ b/src/Helper/Helper_Field_Container.php
@@ -162,7 +162,9 @@ class Helper_Field_Container {
 	public function generate( GF_Field $field ) {
 
 		/* Check if we are processing a field that should not be floated and treat it as a 100% field */
-		$this->process_skipped_fields( $field );
+		if( $this->process_skipped_fields( $field ) ) {
+			return;
+		}
 
 		/* Check if we need to close the container */
 		if ( $this->currently_open ) {


jakejackson1:
Thanks for the contribution @FPCSJames! I'll take a closer look at this issue and your solution next week when I'm back in the office.

jakejackson1:
@FPCSJames I've reviewed this issue and PR in full and will be tagging as `wontfix`. The two main reasons for this are:

1. The skipped fields are related to column support. We are essentially disabling column support for the fields marked as "skipped" (not very intuitive naming convention, I know). When a field is marked as "skipped" it's treated as a field with no column support added (i.e 100% width). 
2. The container wrapper is integral to the PDF template design and formatting. 

I'd suggest using cascading CSS to targeting `.gfpdf-field .inner-container` and then override styles for individual fields using `gfpdf-address .inner-container`.

FPCSJames:
Is your assertion, then, that the existing code is doing exactly what it's supposed to be? Is it exactly as designed that you'd check if a field type should be skipped, then do absolutely nothing with the results of that check? Why even bother to return a value from `process_skipped_fields( $field )` if you're never using the return value?

FPCSJames:
I can see in the history where this was introduced to begin with - when you changed `process_skipped_fields` to start calling `strip_field_of_any_classmaps`, you stopped checking the return value of `process_skipped_fields` in `generate`. The real problem as I can see it is that both are important. Indeed, a skipped field shouldn't have the float-width classes, but as a non-floated element, it also shouldn't be wrapped in a `row-separator`. 

It isn't `.inner-container` that's in question here, it's `.row-separator`.

jakejackson1:
@FPCSJames all good points. This code was written four years ago and so I cannot say with certainty what the intent was at the time. Looking at it now, what I can say is this functionality has been in production for over 4 years and we won't be introducing a change that will break backwards compatibility – we have 30,000+ active installations in the wild to consider. 

If _you_ need this functionality to work the way you described, [we could introduce a filter](https://github.com/GravityPDF/gravity-pdf/blob/development/src/View/View_PDF.php#L372) that would allow you to swap out the Helper_Field_Container class for one of your own. Alternatively, you [could disable the Core template's `enable_css_ready_classes` configuration option](https://github.com/GravityPDF/gravity-pdf/blob/development/src/templates/zadani.php#L248) to prevent the `row-separator` wrapper altogether.

FPCSJames:
Introducing a filter at the point linked wouldn't help in my case - I need it to skip generation of `row-separator` only for skipped field types, and it looks like filtering `$container` in View_PDF.php would change the behavior for the entire entry.

What would help is if the filter simply toggled on the behavior I've described here. I've updated my patch branch with a possible solution - "fixed" behavior is off by default. [See this commit](https://github.com/FPCSJames/gravity-pdf/commit/612284ebdfbbf1d6c7d036b6548ce2054f64b7d6) - not sure if I can add it to the PR now.

jakejackson1:
@FPCSJames you can extend the existing `Helper_Field_Container` class and tweak the logic you need to, then inject your new class in the new filter I mentioned. Saves you duplicating the entire class logic. This is a similar pattern to what we do for processing Gravity Form fields and the one I'm inclined to implement.

Along with the new filter, we'll need to update the `private` methods to `protected` in `Helper_Field_Container` to allow the class to be more extensible (but this should have been the case from the get go).

FPCSJames:
Once you set the private properties to protected, that should help. I'll look into it - thanks.

jakejackson1:
@FPCSJames when you have time please give https://github.com/GravityPDF/gravity-pdf/pull/1051 a go and let me know if the changes will be suitable for you.

PR #1051, opened 2020-03-11
Allow Field_Container class to be more extensible

jakejackson1:
## Description

This PR makes the following changes:
- Set all `private` properties to `protected`
- Add new filter `gfpdf_field_container_class`

Refer to  https://github.com/GravityPDF/gravity-pdf/pull/1048 for the reason(s) why this change came about.

Devs who want to change how the `Helper_Field_Container` class works can now use the filter `gfpdf_field_container_class` to inject their own class (either an entirely new class, or one that extends the existing one and tweaks the functionality). 

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/Helper/Fields/Field_Form.php
+++ b/src/Helper/Fields/Field_Form.php
@@ -2,14 +2,12 @@
 
 namespace GFPDF\Helper\Fields;
 
+use Exception;
+use GFPDF\Helper\Helper_Abstract_Fields;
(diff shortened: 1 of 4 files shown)


jakejackson1:
@FPCSJames this branch should have the changes you need to get the results you're looking for. If you have time, check it out and let me know if any adjustments are needed.

FPCSJames:
@jakejackson1 I have tested this and it looks fine.

PR #1055, opened 2020-04-14
Initial refactor unit test file

jestonihpi:
## Description
Create new directory and Renamed test-action.php to ActionTest.php.
Unload WP_UnitTestCase and define it as \WP_UnitTestCase on ActionTest class.

## Checklist:
- [ ] I've tested the code.


tests/phpunit/unit-tests/Controller/ActionTest.php
```diff
@@ -1,14 +1,11 @@
 <?php
 
-namespace GFPDF\Tests;
+namespace GFPDF\Tests\Controller;
```
jakejackson1, line 3: Ideally, we should use the same namespace as the class being tested.

composer.lock
```diff
@@ -475,16 +475,16 @@
         },
         {
             "name": "psr/log",
-            "version": "1.1.2",
```
jakejackson1, line 478: Let's remove this whole file from the commit since it's not necessary for your PR.

tests/phpunit/unit-tests/Controller/ActionTest.php
```diff
@@ -247,7 +245,7 @@ function( $routes ) {
 		/* Fail capability check */
 		try {
 			$this->controller->route();
-		} catch ( Exception $e ) {
+		} catch ( \Exception $e ) {
```
jakejackson1, line 248: We should look at utilising PHPUnit's `@expectException` instead of running the test this way.  https://phpunit.de/manual/6.5/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.exceptions

jakejackson1 (changes requested):
A number of the test methods are quite verbose. As part of the refactor we should split these larger tests down into smaller test methods.

tests/phpunit/unit-tests/Controller/ActionTest.php
jestonihpi: Noted @jakejackson1

composer.lock
jestonihpi: Fixed on recent push.

tests/phpunit/unit-tests/Controller/ActionTest.php
jestonihpi: Fixed on recent push.

tests/phpunit/unit-tests/Controller/ActionTest.php
jestonihpi: Applied and use WPDieException

.gitignore
```diff
@@ -43,6 +43,7 @@ dist/*
 _notes
 _notes/*
 Thumbs.db
+composer.lock
```
jakejackson1, line 46: The lock file is important to ensure the built copy of Gravity PDF exactly matches what we run in development. My previous comment about not including it in your PR was because it was unrelated to the work you are doing. We try keep our PRs to one specific focus. In this case, refactoring the unit tests.

tests/phpunit/unit-tests/Controller/ActionTest.php
```diff
@@ -286,6 +220,20 @@ function( $routes ) {
...
+	 * @expectedException WPDieException
+	 */
+	public function test_route_exception_permission() {
+		/* Set test routes */
+		$this->set_test_routes();
+
+		$_POST['gfpdf_action'] = 'gfpdf_test_action';
+
+		$this->controller->route();
+
```
jakejackson1, line 233: New lines here that aren't needed.

jakejackson1 (changes requested):
Just a quick review right now. Will do a full one tomorrow.

.gitignore
jestonihpi: Noted Jake.

tests/phpunit/unit-tests/Controller/ActionTest.php
jestonihpi: Fixed.

PR #1056, opened 2020-05-06
Resolve PHP Notice for non-likert survey fields

jakejackson1:
## Description

Changing the type of a Survey Likert field can cause a PHP notice when viewing the PDF.

## Testing instructions
1. Install Survey Add-on
1. Add Survey field to form and save
1. Change survey field from Likert to Radio and save
1. View core PDF with Xdebug enabled

## Checklist:
- [ ] I've tested the code.


jestonihpi:
Tested and working on my end.

PR #1090, opened 2020-08-11
Always parse Core Font payload as JSON

### LINKED ISSUE #1069: Cannot run Core Installer when `content-type` header not returned from API request
When using `superagent` to get `dist/payload/core-fonts.json`, if the response doesn't include a `content-type` header then `superagent` will default to parsing using XML instead of JSON. This causes an error.

jakejackson1:
## Description

Regardless of the `Content-Type` header sent back when accessing the `/dist/payload/core-fonts.json` file, we should always parse a successful API request as JSON. 

Resolves #1069

## Testing instructions

1. Open the `.htaccess` file at the root of your WordPress directory and add `AddType none .json` to the top
2. In the `gf-2.5` branch if you run the Core Font Installer now you will get an error
3. If you checkout this branch and rerun the installer it will function as expected
4. If you remove the line added to `.htaccess` it will still function correctly

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/assets/js/react/api/coreFonts.js
+++ b/src/assets/js/react/api/coreFonts.js
@@ -19,6 +19,10 @@ export function apiGetFilesFromGitHub () {
     .get(GFPDF.pluginUrl + 'dist/payload/core-fonts.json')
     .accept('application/json')
     .type('json')
+    .buffer(true)
+    .parse(function(response) {
+      return JSON.parse(response.text)
+    })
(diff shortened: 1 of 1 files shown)


jakejackson1:
@debbieellis7 could you check this work over when you've time and verify there's no issues?

debbieellis7 (approved):

PR #1106, opened 2020-08-18
Allow class_map to be filtered

JacobDB:
## Description

I ran in to a situation where the built-in Gravity Forms columns classes weren't fulfilling my needs, so I instead wrote my own classes and styled them manually. This created a problem with Gravity PDF, as the column system is entirely reliant on the native Gravity Forms classes, and wraps each field in a `row-separator` element without any custom classes to target. The way you have `$class_map` set up is pretty straightforward for adding additional classes, so I modified it so that it can be filtered to do so.

## Testing instructions
Add custom classes to a form element, for example `custom-50`, and then apply a filter to add that class to the array.

```
function my_class_map($class_map) {
    $class_map["custom-50"] = 50;

    return $class_map;
}
add_filter("my_class_map", "gfpdf_field_width_class_map");
```

## Screenshots 

N/A

## Checklist:
- [x] I've tested the code.

## Additional Comments 

It seems like this may not be the most efficient way to do this, as it calls the filter on every field rather than once for all fields. I don't often work with classes (something I definitely need to study up on), so my initial thought was to simply apply the filter directly to the `$class_map` variable. This resulted in an error "Constant expression contains invalid operations," which I infer means that a function can't be run against a `private` variable like this. Instead, I just applied the filter each time `$this->class_map` is called, which works perfectly, but if you have a more DRY/performant method to achieve this I'm open to suggestions.


jakejackson1:
Thanks for the contribution @JacobDB! When I've time I'll take a closer look at this and see what the best approach is.

jakejackson1:
Hey @JacobDB, appologies for not being able to look in on this sooner. Have you had a look at Gravity Forms 2.5-beta yet? They completely revamp how columns are done in the form editor and may mean you don't need to roll your own custom classes for the PDF in future.

As for a performant way to achieve what you want now, I'd suggest we look at filtering the $class_map property once in the class `__construct()` here: https://github.com/GravityPDF/gravity-pdf/blob/master/src/Helper/Helper_Field_Container.php#L149. What way the filter is running every single time a field is output in the PDF. Thoughts?

JacobDB:
Good call, I've made that adjustment. I haven't looked at that beta, I don't tend to use Gravity Forms very often. Sounds like it should make things a bit easier in the future though!

jakejackson1:
@JacobDB this has been merged into Gravity PDF 5.3.4, however has been renamed to `gfpdf_container_class_map` to match the naming conversion of the other filters in the class.

https://github.com/GravityPDF/gravity-pdf/pull/1186/commits/44c4a5e36d255cececb93563014056d87e1a158a

PR #1142, opened 2020-10-21
WCAG 2.1 Development

jestonihpi:
R&D for WCAG 2.1 Implementation. The main goal was to follow the best practices as close as possible. 
## Description

Added aria tags.

## Checklist:
- [ ] I've tested the code.


src/Helper/Helper_Abstract_Options.php
```diff
@@ -1545,10 +1546,13 @@ public function multicheck_callback( $args ) {
...
+				$aria_checked = !empty( $enabled ) ? 'aria-checked="true"' : 'aria-checked="false"';
+				$aria_label = str_replace( '_', ' ',$label );
 
 				echo '<div class="gfpdf-settings-multicheck-wrapper">';
-				echo '<input name="gfpdf_settings[' . $args['id'] . '][' . $args['multi-key'] . ']" id="gfpdf_settings[' . $args['id'] . '][' . $args['multi-key'] . ']" class="gfpdf_settings_' . $args['id'] . ' ' . $class . '" type="checkbox" value="' . $args['multi-key'] . '" ' . checked( $key, $enabled, false ) . ' ' . $required . ' />&nbsp;';
-				echo '<label for="gfpdf_settings[' . $args['id'] . '][' . $args['multi-key'] . ']">' . esc_html( $option ) . '</label>';
+				echo '<input name="gfpdf_settings[' . $args['id'] . '][' . $args['multi-key'] . ']" id="gfpdf_settings[' . $args['id'] . '][' . $args['multi-key'] . ']" class="gfpdf_settings_' . $args['id'] . ' ' . $class . '" type="checkbox" value="' . $args['multi-key'] . '" ' . checked( $key, $enabled, false ) . ' ' . $required . ' role="checkbox" ' . $aria_checked . ' />&nbsp;';
+				echo '<label for="gfpdf_settings[' . $args['id'] . '][' . $args['multi-key'] . ']" aria-label="'. $aria_label .'">' . $label . '</label>';
```
jakejackson1, line 1555: All my reading suggests that `aria` tags aren't necessary on a `<label>` tag with a correctly referenced `for` target:

https://stackoverflow.com/questions/55814279/aria-label-vs-form-labels
https://www.w3.org/TR/WCAG20-TECHS/ARIA14.html
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute

Could you please provide reference martial that backs up why using both is appropriate?

src/Helper/Helper_Abstract_Options.php
```diff
@@ -1545,10 +1546,13 @@ public function multicheck_callback( $args ) {
 				$args['multi-option'] = $key;
 
 				$enabled = $this->get_form_value( $args );
+				$label = esc_html( $option ) ;
+				$aria_checked = !empty( $enabled ) ? 'aria-checked="true"' : 'aria-checked="false"';
+				$aria_label = str_replace( '_', ' ',$label );
```
jakejackson1, line 1551: This code needs to be linted

src/Helper/Helper_Abstract_Options.php
```diff
@@ -1545,10 +1546,13 @@ public function multicheck_callback( $args ) {
...
 
 				echo '<div class="gfpdf-settings-multicheck-wrapper">';
-				echo '<input name="gfpdf_settings[' . $args['id'] . '][' . $args['multi-key'] . ']" id="gfpdf_settings[' . $args['id'] . '][' . $args['multi-key'] . ']" class="gfpdf_settings_' . $args['id'] . ' ' . $class . '" type="checkbox" value="' . $args['multi-key'] . '" ' . checked( $key, $enabled, false ) . ' ' . $required . ' />&nbsp;';
-				echo '<label for="gfpdf_settings[' . $args['id'] . '][' . $args['multi-key'] . ']">' . esc_html( $option ) . '</label>';
+				echo '<input name="gfpdf_settings[' . $args['id'] . '][' . $args['multi-key'] . ']" id="gfpdf_settings[' . $args['id'] . '][' . $args['multi-key'] . ']" class="gfpdf_settings_' . $args['id'] . ' ' . $class . '" type="checkbox" value="' . $args['multi-key'] . '" ' . checked( $key, $enabled, false ) . ' ' . $required . ' role="checkbox" ' . $aria_checked . ' />&nbsp;';
```
jakejackson1, line 1554: This is an invalid use of `aria-checked`. See:

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role

src/Helper/Helper_Options_Fields.php
```diff
@@ -237,7 +236,7 @@ public function get_registered_fields() {
 				'uninstaller' => [
 					'id'         => 'uninstaller',
 					'name'       => esc_html__( 'Uninstall Gravity PDF', 'gravity-forms-pdf-extended' ),
-					'desc'       => '<p class="alert error">' . esc_html__( 'This operation deletes ALL Gravity PDF settings and deactivates the plugin. If you continue, all settings, configuration, custom templates and fonts will be removed.', 'gravity-forms-pdf-extended' ) . '</p>',
+					'desc'       => '<p class="alert error" role="dialog">' . esc_html__( 'This operation deletes ALL Gravity PDF settings and deactivates the plugin. If you continue, all settings, configuration, custom templates and fonts will be removed.', 'gravity-forms-pdf-extended' ) . '</p>',
```
jakejackson1, line 239: The use of `role="dialog"` isn't valid in this context:

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role

It's only used for dialog boxes, but this message displays right above the button it relates to.

src/Helper/Helper_PDF_List_Table.php
```diff
@@ -228,6 +228,8 @@ public function column_cb( $item ) {
 			 data-fid="<?= $form_id ?>"
 			 src="<?= $this->gform->get_plugin_url() ?>/images/active<?= intval( $is_active ) ?>.svg"
 			 alt="<?= $label ?>"
+			 role="img"
+			 aria-label="<?= $label ?>"
```
jakejackson1, line 232: `role="img"` definitely isn't the right approach here as it's assigned to a single image and not a container with a "group" of images https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Role_Img

. `arial-label` not correct either. Likely need aria-live on a new visible element: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions. Since GF has updated how the toggle looks this should be much easier to achieve.

jakejackson1 (changes requested):
Along with me notes, you need to rebase this branch on the latest `development` branch and then force push.

src/Helper/Helper_Abstract_Options.php
jestonihpi: As for my test Jake. aria-label is prioritized by screen reader. My goal was to design the labels to more user friendly in terms of audio. Thus $aria_label = str_replace( '_', ' ',$label ); e.g gravityforms_view_entries will be read as gravityforms view entries.
Anyways this item is still experimental.

src/Helper/Helper_Options_Fields.php
jestonihpi: This will be replaced with role="alert".

jakejackson1: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role

This would suggest that the alert role should be used for validation messages and not for generate descriptions about an action. It seems that `aria-describedby` would be more practical here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute

src/Helper/Helper_Abstract_Options.php
jakejackson1: That makes sense, but we are talking about specific capability names [which are described in detail in the Gravity Forms documentation](https://docs.gravityforms.com/role-management-guide/). If they had read the GF docs, users would be looking for `gravityforms*underscore*view*underscore*entries`, rather than `gravityforms view entries`.

src/assets/js/react/components/FontManager/SearchBox.js
```diff
@@ -120,17 +120,19 @@ export class SearchBox extends Component {
...
-        id='font-manager-search-box'
-        className='wp-filter-search'
-        placeholder={GFPDF.fontManagerSearchPlaceHolder}
-        value={this.state.searchInput}
-        onChange={this.handleSearch}
-        onKeyDown={e => e.keyCode === 13 && e.preventDefault()}
-        ref={node => (this.input = node)}
-        tabIndex='143'
-      />
+      <div role='form'>
```
jakejackson1, line 123: Couple comments on this:

1. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Form_Role recommends using `<form>` where possible (but doesn't go into why you would choose to use `role="form"` instead of `<form>`). What research have you done into use the role-base alternative, over using the native `<form>` tag?

2. Regardless of which way we go, shouldn't we look at wrapping both the search and results in a FORM element, since the results include form-related actions (delete for example)?

jakejackson1: Also, it looks like `<form role="search">` is preferred https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Search_role

jakejackson1:
@jestonihpi please rebase your PR on the latest `development` branch and only include changes relevant to your work (i.e no code reformatting).

jakejackson1:
Closed in favour of https://github.com/GravityPDF/gravity-pdf/pull/1177

Issue #1112, opened 2020-08-20
Double Ajax Requests after uploading a new template and selecting it


jakejackson1:
To replicate:

1. Adding a new PDF to a form
1. Go to the Template Manager and upload a new zip file
1. Select the newly installed template
1. Watch as two loading spinners are displayed

This usually results in the incorrect template and template settings being loaded. We need to fix this issue.

jakejackson1:
@debbieellis7 was able to replicate in the 6.0 branch. Here's a Loom: https://www.loom.com/share/7ed2cf7c784a4929ab6347ef63b9596b

PR #1184, opened 2021-01-25
Fixed template double spinner bug

debbieellis7:
## Description

Issue: #1112 
Fixed double Ajax Requests after uploading a new template and selecting it.

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/assets/js/react/bootstrap/templateBootstrap.js
+++ b/src/assets/js/react/bootstrap/templateBootstrap.js
@@ -91,15 +91,15 @@ export function activeTemplateStoreListener (store, $templateField) {
   $templateField.on('change', function () {
     /* Check store and DOM are different to prevent any update recursions */
     if (this.value !== store.getState().template.activeTemplate) {
-      store.dispatch(selectTemplate(this.value))
+      store.dispatch(selectTemplate(store.getState().template.activeTemplate))
     }
   })
 }
 
 /**
- * PHP builds the Select box DOM for the templates and when we add or delete a template we need to rebuild this.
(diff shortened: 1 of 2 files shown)


src/assets/js/admin/settings/common/setupDynamicTemplateFields.js
```diff
@@ -15,6 +15,13 @@ import { toggleFontAppearance } from '../pdf/toggleFontAppearance'
 export function setupDynamicTemplateFields () {
   /* Add change listener to our template */
   $('#gfpdf_settings\\[template\\]').off('change').on('change', function () {
+    const existingSpinner = document.querySelector('.gfpdf-spinner-template')
+
+    /* Remove existing spinner to avoid duplication */
```
jakejackson1, line 20: @debbieellis7 This seems like it's only masking the problem and not actually stopping the double AJAX request. That double request would still be made, but only one spinner would be shown.

debbieellis7: Yes, This will do for now @jakejackson1. We need to convert to react so it can access our redux store and we will need a discussion for that one first.

jakejackson1: Instead of removing the existing spinner, can you just return early and not do anything if there's already a request in progress?

debbieellis7: Yes, I think it will work. Kindly check the latest pushed code @jakejackson1. Thanks

Issue #1195, opened 2021-03-01
Font Manager Font List Radio Selector Not Accessible


jakejackson1:
**Description**
These inputs are not accessible via keyboard navigation, and it doesn't include an ARIA description about what it does.

**Step To Reproduce**
1. Go to Settings -> PDFs
2. Click on Advanced button next to the font selector
3. Tab through the List

**Expected behavior**
To be able to tab through each font in the list and mark it as selected. Should also receive screenreader output on the radio field selector explaining what it is. 

**Screenshots**
<img width="553" alt="Screen Shot 2021-03-01 at 2 55 57 pm" src="https://user-images.githubusercontent.com/2918419/109450120-453d5180-7a9e-11eb-80a0-38c6ca213473.png">

PR #1205, opened 2021-03-02
Fix font list radio button keyboard accessibility bug

debbieellis7:
## Description

Issue: #1195 

To be able to tab through each font in the list and mark it as selected. Should also receive screenreader output on the radio field selector explaining what it is.

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/assets/js/react/components/FontManager/FontListItems.js
+++ b/src/assets/js/react/components/FontManager/FontListItems.js
@@ -286,6 +286,19 @@ export class FontListItems extends Component {
    */
   handleSelectFont = e => {
     this.props.selectFont(e.target.value)
+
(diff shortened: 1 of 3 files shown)


src/assets/js/react/components/FontManager/FontListItems.js
```diff
@@ -337,8 +350,8 @@ export class FontListItems extends Component {
                 {!disableSelectFontName && (
                   <input
                     type='radio'
-                    className='selectFontName'
-                    name='selectFontName'
+                    className='select-font-name'
+                    name={'select-font-name-' + font.id}
```
jestonihpi, line 354: Add an aria label that describe what does the button do. e.g aria-label="Select {Font name} font".

PR #1218, opened 2021-03-09
System status missing clipboard data and default values.

jestonihpi:
## Description
Set default value on status page in which solves the problem with empty data on clipboard.

## Checklist:
- [x] I've tested the code.


jakejackson1 (changes requested):
@jestonihpi Background Processing and Debug Mode are using the wrong fallback value.

jestonihpi:
Updated fall back values @jakejackson1 .

jakejackson1:
> Updated fall back values @jakejackson1 .

`'background_processing'` and `'debug_mode'` still have incorrect default values. Just lookup an existing call to these options and copy the default we're already using. Don't guess!

PR #1234, opened 2021-03-21
Keyboard navigation for mergetag selector

### LINKED ISSUE #1226: Cannot tab to Mergetag Selector
**Description**
On the PDF creation page, when tabbing through the UI I am unable to tab directly to the Mergetag selector and navigate through the component using my keyboard only. 

**Step To Reproduce**
1. Go to PDF List page and add a new PDF
2. Fill in the Name setting then tab through the components until you get to the Filename
3. When you next tab you should go to the Mergetag selector, but do not.

**Expected behavior**
To be able to navigate and interact with the UI via the keyboard alone.

**Screenshots**
Screen Shot 2021-03-14 at 8 24 31 pm

debbieellis7:
## Description
Issue: #1226 

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/assets/js/react/gfpdf-main.js
+++ b/src/assets/js/react/gfpdf-main.js
@@ -5,6 +5,7 @@ import coreFontBootstrap from './bootstrap/coreFontBootstrap'
 import helpBootstrap from './bootstrap/helpBootstrap'
 import addEditButton from './utilities/PdfSettings/addEditButton'
 import autoSelectShortcode from './utilities/PdfList/autoSelectShortcode'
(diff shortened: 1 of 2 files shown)


jakejackson1:
@debbieellis7 I've been able to tab to the selector and open it with enter, then close it with escape. But when it's open, I cannot tab into any of the available options and select them with only the keyboard. Does this work on your end?

jakejackson1:
Closing due to not being fully keyboard accessible to select tags. As this is a Gravity Forms UI element, let's report it upstream and see if they can fix it wholesale.

PR #1259, opened 2021-04-07
Optimize column widths in PDFs

### LINKED ISSUE #1258: Multiple columns on bordered core templates doesn't fit.
**Description**
Gravity (3) Columns doesn't fit with Zadani and Rubix pdf layout on both RTL and LTR views. 

**Step To Reproduce**
1. Go to 'Forms'
2. Use the core templates Zadani and Rubix.
3. View PDF

**Expected behavior**
Multiple columns length should fit along side with single column.

**Screenshots**
image
image
image
image
4 columns.
image

jakejackson1:
## Description

Force the last column in each row to have a 100% width so the rows can line up better.

Resolves #1258

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/Helper/Helper_Field_Container_Gf25.php
+++ b/src/Helper/Helper_Field_Container_Gf25.php
@@ -101,6 +101,31 @@ class Helper_Field_Container_Gf25 extends Helper_Field_Container {
 	}
 
 	/**
+	 * Close the current container if still open.
+	 * This is usually called publicly after the form loop
+	 *
+	 * @since 6.0
+	 */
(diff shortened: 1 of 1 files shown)


jakejackson1:
@jestonihpi inspiration struck as I thought about how we could provide a better user experience in a way that was going to be compatible with the new way to handle columns. I think I might have cracked it.

jestonihpi:
Copy @jakejackson1 .

jestonihpi (approved):
This updates works as intended on both Zadani and Focus core template.

PR #1310, opened 2022-01-30
Add a custom class on #gfpdf_pdf_form to get the current GF version a…

### LINKED ISSUE #1308: Merge tag ICON needs an update.
**Description**
Inline with GF 2.6-beta's merge tag UI update. We need to update these icons on our end. 

**Step To Reproduce**
1. Go to 'Forms->Settings'
2. Click on 'PDF -> {Sample Template}'
3. Scroll down to 'Notification/Template Settings '

**Expected behavior**
All merge-tag icons under the Gravity PDF roof should set a pointer once hovered.  Larger icon size (50%) for RTEs is suggested.
Updates : The RTE's width should fit to its container as well:
image

Initial Fix:
image

**Screenshots**
image

Note :
This style should be only applied to GF 2.6.

jestonihpi:
…nd apply the request style for 2.6.beta.

## Description
On add-edit.php on FormSettings View. I have added a class which reflects the current version GravityForms. I have prepared that value into a viable class name, then apply the requested styles for the merge tag selector icon and RTE width.

#1308 
## Testing instructions
Make sure your on GF 2.6.beta and WP 5.9, go to Forms->Settings->PDF. The larger merge tag icon and the RTE updated width should be applied.

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/View/html/FormSettings/add_edit.php
+++ b/src/View/html/FormSettings/add_edit.php
@@ -17,7 +17,6 @@ if ( ! defined( 'ABSPATH' ) ) {
 /** @var $args array */
 
 global $wp_settings_fields;
-
 ?>
 
 <!-- Merge tag functionality requires a global form object -->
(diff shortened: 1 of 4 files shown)


jakejackson1:
@jestonihpi what happens when Gravity Forms is updated to 2.7? This code is very brittle. Better to do the version detection in PHP using `version_compare` and then inject a specific CSS class that will apply the new styles.

src/View/html/FormSettings/add_edit.php
```diff
@@ -18,6 +18,11 @@
 
 global $wp_settings_fields;
 
+$additional_form_class = '';
+
+if ( version_compare( '2.6', GFCommon::$version, '>=' ) ) {
```
jakejackson1: I think we need to check `2.6-beta-1` rather than just `2.6` for this to work with the beta correctly. 

Also, this logic should be moved to the Controller or Model (depending who calls `add_edit`) and then injected into the view.

jakejackson1: We should probably use a class name that isn't so generic either like `gfpdf-gf-2-6`

jakejackson1:
@jestonihpi looks good. Please do a rebase on the latest `origin/development` branch and force push your changes again. I'll then review and then get this merged.

jestonihpi:
Done @jakejackson1. Thanks!

mcglenn07 (commented):
Updated RTE width and mergetag icon now displaying properly.

PR #1317, opened 2022-02-28
Add Zapier Support (PDF URLS)

### LINKED ISSUE #1287: Add support for new Zapier workflow
**Description**
Zapier is now fully setup in Zapier and not in Gravity Forms. We need to ensure the PDF URLs are passed along and available for processing

jestonihpi:
## Description
Since the introduction of Zapier filter `gform_zapier_request_body`  (3.2). We will be including the PDF URLs on its returned `$body`. 

Fixes #1287 
## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/Controller/Controller_Zapier.php
+++ b/src/Controller/Controller_Zapier.php
@@ -1,0 +1,58 @@
+<?php
+
+namespace GFPDF\Controller;
+
+/**
+ * @package     Gravity PDF
+ * @copyright   Copyright (c) 2022, Blue Liquid Designs
+ * @license     http://opensource.org/licenses/gpl-2.0.php GNU Public License
+ */
+
+/* Exit if accessed directly */
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Class Controller_Zapier
+ *
+ * @package GFPDF\Controller
+ */
+class Controller_Zapier {
+
+	/**
+	 * @since 6.3
+	 */
+	public function init(): void {
+		add_filter( 'gform_zapier_request_body', [ $this, 'add_zapier_support' ], 10, 3 );
+	}
+
+	/**
+	 * Add PDF URLs to the Zapier body array
+	 *
+	 * @param array $body  An associative array containing the request body that will be sent to Zapier.
+	 * @param array $feed  The Feed Object currently being processed.
+	 * @param array $entry The Entry Object currently being processed.
+	 *
+	 * @return array
+	 *
+	 * @since 6.3
+	 */
+	public function add_zapier_support( $body, $feed, $entry ) {
+		$zapier = \GF_Zapier::get_instance();
+		$pdfs   = \GPDFAPI::get_entry_pdfs( $entry['id'] ?? 0 );
+
+		if ( is_wp_error( $pdfs ) ) {
+			return $body;
+		}
+
+		foreach ( $pdfs as $pdf ) {
+			$body[ $zapier->get_body_key( $body, $pdf['name'] . ' PDF URL' ) ]                  = do_shortcode( sprintf( '[gravitypdf id="%2$s" entry="%1$d" raw="1"]', $entry['id'], $pdf['id'] ) );
+			$body[ $zapier->get_body_key( $body, $pdf['name'] . ' PDF URL - SIGNED 1 WEEK' ) ]  = do_shortcode( sprintf( '[gravitypdf id="%2$s" entry="%1$d" raw="1" signed="1" expires="+1 week"]', $entry['id'], $pdf['id'] ) );
+			$body[ $zapier->get_body_key( $body, $pdf['name'] . ' PDF URL - SIGNED 1 MONTH' ) ] = do_shortcode( sprintf( '[gravitypdf id="%2$s" entry="%1$d" raw="1" signed="1" expires="+1 month"]', $entry['id'], $pdf['id'] ) );
+			$body[ $zapier->get_body_key( $body, $pdf['name'] . ' PDF URL - SIGNED 1 YEAR' ) ]  = do_shortcode( sprintf( '[gravitypdf id="%2$s" entry="%1$d" raw="1" signed="1" expires="+1 year"]', $entry['id'], $pdf['id'] ) );
+		}
+
+		return $body;
+	}
+}
--- a/src/bootstrap.php
+++ b/src/bootstrap.php
@@ -226,6 +226,7 @@ class Router implements Helper\Helper_Interface_Actions, Helper\Helper_Interface
 		$this->check_system_status();
 		$this->export();
 		$this->webhooks();
+		$this->zapier();
 
 		/* Add localisation support */
 		$this->add_localization_support();
@@ -894,6 +895,16 @@ class Router implements Helper\Helper_Interface_Actions, Helper\Helper_Interface
 	 */
 	public function webhooks(): void {
 		$class = new Controller\Controller_Webhooks();
+		$class->init();
+
+		$this->singleton->add_class( $class );
+	}
+
+	/**
+	 * @since 6.3
+	 */
+	public function zapier(): void {
+		$class = new Controller\Controller_Zapier();
 		$class->init();
 
 		$this->singleton->add_class( $class );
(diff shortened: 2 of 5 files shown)


src/bootstrap.php
```diff
@@ -355,6 +358,31 @@ public function add_body_class( $classes ) {
...
+	 * @since 6.1
+	 */
+	public function add_expanded_zapier_support( $body, $feed, $entry, $form ) {
+		$zapier = \GF_Zapier::get_instance();
+		$signer = new Helper\Helper_Url_Signer();
+		$model  = new Model\Model_PDF( $this->gform, $this->log, $this->options, $this->data, $this->misc, $this->notices, $this->templates, $signer );
+
+		/* @todo : Signed urls */
+		foreach ( $form['gfpdf_form_settings'] as $settings ) {
+			$body[ $zapier->get_body_key( $body, $settings['template'] . ' PDF URL' ) ] = $model->get_pdf_url( $settings['id'], $entry['id'] );
```
jakejackson1: This should be `$settings['label']` instead of `$settings['template']` I think.

src/bootstrap.php
```diff
@@ -355,6 +358,31 @@ public function add_body_class( $classes ) {
...
+	 * @param array $form
+	 *
+	 * @return array
+	 *
+	 * @since 6.1
+	 */
+	public function add_expanded_zapier_support( $body, $feed, $entry, $form ) {
+		$zapier = \GF_Zapier::get_instance();
+		$signer = new Helper\Helper_Url_Signer();
+		$model  = new Model\Model_PDF( $this->gform, $this->log, $this->options, $this->data, $this->misc, $this->notices, $this->templates, $signer );
```
jakejackson1: You can change this to `\GPDFAPI::get_mvc_class( 'Model_PDF' )`

src/bootstrap.php
jakejackson1: The fifth property passed to `get_pdf_url()` should be set to `false`

jakejackson1: To add signed URL support you would duplicate this line, adjust the body key, and then wrap the PDF URL in `$signer->sign( $url, '+1 week' )`.

Repeat this process for both a month and year: `$signer->sign( $url, '+1 month' )` or `$signer->sign( $url, '+1 year' )`.

jakejackson1 (changes requested):
All this code needs to be moved to a new controller file that is only initialised if the `GF_Zapier` class exists.

jakejackson1:
@jestonihpi Unit tests will also need to be written. We need to check what happens if a PDF is active, inactive, conditional logic doesn't pass for the current entry, or there isn't any PDFs configured on the form at all and thus no `$form['gfpdf_form_settings']` key. Each scenario will change the results of this method.

src/bootstrap.php
jestonihpi: Copy @jakejackson1

src/Controller/Controller_Zapier.php
```diff
@@ -38,22 +39,35 @@ public function init(): void {
...
-	 * @throws \GFPDF_Vendor\Spatie\UrlSigner\Exceptions\InvalidExpiration
-	 * @throws \GFPDF_Vendor\Spatie\UrlSigner\Exceptions\InvalidSignatureKey
+	 * @throws \Exception
 	 * @since 6.1
 	 */
 	public function add_expanded_zapier_support( $body, $feed, $entry, $form ) {
 		$zapier = \GF_Zapier::get_instance();
 		$signer = new Helper_Url_Signer();
 		$model  = \GPDFAPI::get_mvc_class( 'Model_PDF' );
+		$logger = new Helper_Logger( 'gravity-pdf', 'Gravity PDF' );
```
jakejackson1: You need to pass `$log` from the bootstrap file to this class's `__construct()`.

src/Controller/Controller_Zapier.php
```diff
@@ -38,22 +39,35 @@ public function init(): void {
...
+			} catch ( \Exception $e ) {
+				$log->error(
+					'Generating Zapier signed PDF URLs error.',
+					[
+						'entry'     => $entry,
+						'settings'  => $settings,
+						'exception' => $e->getMessage(),
+					]
+				);
+				throw new \Exception( 'Zapier signed PDF URLs error.' );
```
jakejackson1: Remove this line. You just caught the Exceptions and handled them appropriately. No need to throw a new error.

jakejackson1 (changes requested):

jestonihpi:
Hi @jakejackson1 . Can you review the recent update for this PR as well? Thank you!

jakejackson1:
@jestonihpi didn't you mentioned you were working on unit tests for this PR?

The `@since` docblocks all need to be updated to `6.3`

src/Controller/Controller_Zapier.php
```diff
@@ -0,0 +1,86 @@
...
+	/**
+	 * Controller_Zapier constructor.
+	 *
+	 * Setup our class by injecting all our dependencies
+	 *
+	 * @param Helper_Logger $log
+	 *
+	 * @since 6.3
+	 */
+	public function __construct( $log ) {
```
jakejackson1: Need to type hint `Helper_Logger` here

src/Controller/Controller_Zapier.php
```diff
@@ -0,0 +1,86 @@
...
+		/* Load Logger */
+		$this->log = $log;
```
jakejackson1: Cannot see a property called `log` defined in this class

src/Controller/Controller_Zapier.php
```diff
@@ -0,0 +1,86 @@
...
+
+	}
+	/**
```
jakejackson1, line 23: Missing new line before this docblock

src/Controller/Controller_Zapier.php
```diff
@@ -0,0 +1,86 @@
...
+	 * Add PDF URLs to Zapier's data(body).
+	 *
+	 * @param array $body
+	 * @param array $feed
+	 * @param array $entry
+	 * @param array $form
+	 *
+	 * @return array
+	 *
+	 * @throws \Exception
```
jakejackson1: This no longer throws an exception

src/Controller/Controller_Zapier.php
```diff
@@ -0,0 +1,86 @@
...
+	 * @since 6.0
+	 */
+	public function init(): void {
+		add_filter( 'gform_zapier_request_body', [ $this, 'add_expanded_zapier_support' ], 10, 4 );
+	}
+
+	/**
+	 * Add PDF URLs to Zapier's data(body).
+	 *
+	 * @param array $body
```
jakejackson1: A brief description of each param is useful

src/Controller/Controller_Zapier.php
```diff
@@ -0,0 +1,86 @@
...
+	}
+	/**
+	 * @since 6.0
+	 */
+	public function init(): void {
+		add_filter( 'gform_zapier_request_body', [ $this, 'add_expanded_zapier_support' ], 10, 4 );
+	}
+
+	/**
+	 * Add PDF URLs to Zapier's data(body).
```
jakejackson1: add space before `(`

src/bootstrap.php
```diff
@@ -226,6 +226,7 @@ public function init() {
 		$this->check_system_status();
 		$this->export();
 		$this->webhooks();
+		$this->zapier( $this->log );
```
jakejackson1: You don't need to pass $log around here. It's accessible throughout the entire bootstrap.php file

src/bootstrap.php
```diff
@@ -899,6 +900,16 @@ public function webhooks(): void {
 		$this->singleton->add_class( $class );
 	}
 
+	/**
+	 * @since 6.3
+	 */
+	public function zapier( $log ): void {
+		$class = new Controller\Controller_Zapier( $log );
```
jakejackson1: Pass `$this->log` directly.

src/Controller/Controller_Zapier.php
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/**
+ * Class Controller_Zapier
+ *
+ * @package GFPDF\Controller
+ */
+class Controller_Zapier {
+
+	private $log;
```
jakejackson1: @jestonihpi unless we really need to prevent classes that extend this one from accessing the $log property, we use `protected` and not `private`

Issue #1323, opened 2022-03-11
6.2, Revamp on PDF listing row class (gf-locking/alternate)


jestonihpi:
**Description**
We need to re-visit the class changes for the PDF List rows. Previously we added an `if` statement that toggles `alternate` and `gf-locking` depending on the current version of Gravity Forms (`gf-locking` was applied to 2.6-rc-x versions).

**Step To Reproduce**
1. Go to 'Forms->Settings'
2. Click on 'PDF'
3. Scroll down to 'PDF List'
4. Inspect error

**Expected behavior**

**Screenshots**

**Additional context**

jakejackson1:
Fixed https://github.com/GravityPDF/gravity-pdf/pull/1324

PR #1324, opened 2022-03-11
PDF List Table Rows updates.

jestonihpi:
## Description
Remove the conditional classes on each table rows (tr) and create a customize css which implements the same background color of `.alternate`.

#1323 
## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/assets/scss/Pages/_pdf-list.scss
+++ b/src/assets/scss/Pages/_pdf-list.scss
@@ -17,51 +17,58 @@
 }
 
 #tab_PDF {
-
   #gfpdf_list_form {
 
     .gfpdf-spinner {
       width: 20px;
     }
 
-    /* Modify the Gravity Forms 2.6 shortcode button version to fit in our PDF list table. */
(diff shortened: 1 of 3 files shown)


src/assets/scss/Pages/_pdf-list.scss
```diff
@@ -17,51 +17,58 @@
...
-        justify-content: center;
-      }
-    }
-
-    .btn-shortcode + .gpdf-fallback-input {
-      display: none;
-    }
+    table {
+      tbody {
+        & > :nth-child(2n+1) {
```
jakejackson1: No need to overcomplicate this. `td:nth-child(odd)` will give you the same result and be much easier to understand.

jakejackson1 (changes requested):

jestonihpi:
Hi @jakejackson1 . I have updated the recent request to simplify the scss code ( switch to `odd` ). I think this PR is now ready for the final review.

PR #1328, opened 2022-04-14
Conditionally clean up PDF after form submission

### LINKED ISSUE #1320: Do not do form submission PDF cleanup when Background Processing enabled
**Description**

There's a race condition that can occur if the background processor generates the PDF and saves it to disk before the form submission hook is triggered, which is where we have a PDF cleanup routine that gets run.

```
add_action( 'gform_after_submission', [ $this->model, 'cleanup_pdf' ], 9999, 2 );
```

PDFs should not be cleaned up right after form submission if Background Processing is enabled. Let the daily cron cleanup routine handle it instead.

jakejackson1:
## Description

This resolves a race condition with Background Processing.

Fixes #1320

## Testing instructions
1. Set a breakpoint in Xdebug on line 1850 of /src/Model/Model_PDF.php
2. Do form submission and check `Model_PDF::cleanup_pdf()` runs
3. Enable Background Processing
4. Do form submission and check `Model_PDF::cleanup_pdf()` does not run

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/Controller/Controller_Pdf_Queue.php
+++ b/src/Controller/Controller_Pdf_Queue.php
@@ -341,14 +341,23 @@ class Controller_Pdf_Queue extends Helper_Abstract_Controller implements Helper_
 		$queue_data = apply_filters( 'gfpdf_queue_pre_pdf_creation', [], $entry, $form );
 
 		foreach ( $pdfs as $pdf ) {
+			$pdf_queue_data = [
+				'id'            => $this->get_queue_id( $form, $entry, $pdf ),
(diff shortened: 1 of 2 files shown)


jakejackson1:
Resolves https://github.com/GravityPDF/gravity-pdf/issues/1332

jakejackson1:
Resolves https://github.com/GravityPDF/gravity-pdf/issues/1332

jestonihpi (approved):
`Model_PDF::cleanup_pdf()` now triggers correctly and passed through unit-test run (multi).

PR #1330, opened 2022-04-14
Remove stripslashes() when displaying settings

### LINKED ISSUE #1326: Backslash `\` stripped out automatically in PDF Settings after saving
**Description**
The `\` character does get saved to the PDF settings, but when the settings are accessed again this character is getting automatically stripped out of all settings fields.

**Step To Reproduce**
1. Go to Form PDF List page
2. Add new PDF
3. Set the Label to `PDF \ Name`
4. Save
 
**Expected behavior**
Upon saving the backslash character would be displayed on page reload

**Additional context**
You can verify `\` is being saved to the database by navigating back to the PDF List page after the initial save.

jakejackson1:
## Description

Gravity PDF would successfully save settings with a backslash, but when viewed again from the settings page those characters would be automatically stripped out.

Resolves #1326

## Testing instructions
1. Go to a form's PDF settings 
2. Add a backslash in the Label setting
3. Save the settings
4. See that the backslash is still present in the Label on reload  
5. Save the settings 
6. Go to step 4 

~PHP FATAL ERROR: INFINITE LOOP DETECTED~

## Checklist:
- [x] I've tested the code.


jestonihpi (approved):
Confirmed that backslashes are now properly saved on PDF's label.

PR #1340, opened 2022-05-18
Empty repeater field fix

### LINKED ISSUE #1297: Prevent Repeater field showing up in PDF if empty
Prevent empty Repeater fields showing up in the PDF.

jestonihpi:
## Description
This fixes the issue #1297

## Checklist:
- [x] I've tested the code.


jakejackson1 (changes requested):
@jestonihpi can we add a unit test to verify this works as expected?

jestonihpi:
Copy.

jakejackson1:
@jestonihpi this works as expected if the Repeater field is not filled out at all. However, Gravity Forms does allow Repeaters to be Nested together and I'd like to see empty Nested Repeater fields removed from the PDF as well.

Here is an example of this occurring in Focus Gravity: [PDF.pdf](https://github.com/GravityPDF/gravity-pdf/files/9016310/PDF.pdf). The top-level repeater is filled out, but the two nested repeaters are not. 

Form used: [gravityforms-export-2022-06-30.json.txt](https://github.com/GravityPDF/gravity-pdf/files/9016309/gravityforms-export-2022-06-30.json.txt) (remove `.txt` extension from file before importing into Gravity Forms).

jestonihpi:
Added unit test and updated the codebase @jakejackson1 . Thanks.

src/Helper/Fields/Field_Repeater.php
```diff
@@ -187,7 +187,7 @@ public function get_repeater_html( $value, $field ) {
 				if ( $sub_field instanceof GF_Field_Repeater ) {
 
 					/* Only recursively output if a value exists */
-					if ( isset( $item[ $sub_field->id ] ) ) {
+					if ( ! empty( $item[ $sub_field->id ] ) ) {
```
jakejackson1, line 190: Would it make more sense to use `! $field->empty_deep( $item[ $sub_field->id ] )` here, to standardise it like the other empty checks?

src/Helper/Fields/Field_Repeater.php
```diff
@@ -248,5 +248,13 @@ public function value() {
 		$this->cache( $this->get_value() );
 
 		return $this->cache();
+
+	}
+
+	public function maybe_show_section_title( $is_top_level, $value, $field ) {
```
jakejackson1: This probably doesn't need to be `public`.

PR #1341, opened 2022-05-24
Add HTML on Form Data.

### LINKED ISSUE #1300: Repeater: Form_Data doesn't include non-standard fields like HTML/Section
**Description**
When using a Repeater field, the form data only includes fields that would normally be added to the `$form_data['field']` object. Being able to easily access the HTML or Section Break information that are included in Repeaters would be helpful. 

**Step To Reproduce**
1. Create a Repeater field that contains a HTML field
2. Submit a new test entry
3. View the $form_data using the `?data=1` helper parameter

**Expected behavior**
It would be good to see the HTML field content under the `repeater` key.

jestonihpi:
## Description
This update add HTML fields on Form data on Repeater fields.  By adding a `if` statement that checks if the current element is a HTML then manually insert them to the form data.

#1300 

## Testing instructions
Create a sample repeater form with an HTML element on it.  Assign form to any core template and submit a sample entry. Add `?data=1` to the PDF url.

## Checklist:
- [x] I've tested the code.


src/Helper/Fields/Field_Html.php
```diff
@@ -58,11 +58,15 @@ public function __construct( $field, $entry, Helper_Abstract_Form $gform, Helper
 	 */
 	public function form_data() {
 
-		$data = [];
-		$html = $this->value();
-
-		$data['html'][]                      = $html;
-		$data['html_id'][ $this->field->id ] = $html;
+		$data  = [];
```
jakejackson1: @jestonihpi upon thinking about this, I'd rather we didn't modify the `Field_Html.php` file. Let's just merge in the HTML data from the `html_id` key in the Field_Repeater.php like you are already doing.

jakejackson1 (changes requested):
@jestonihpi left you a note

src/Helper/Fields/Field_Html.php
jestonihpi: Copy.

jakejackson1:
@jestonihpi this is missing unit tests.

src/Helper/Fields/Field_Repeater.php
```diff
@@ -107,6 +107,22 @@ public function get_repeater_form_data( $data, $value, $field ) {
...
 					$data = array_replace_recursive( $data, [ $id => $form_data['field'] ] );
 				}
+
+				/* Allow HTML to be included in $form_data */
+				if ( isset( $form_data['html_id'] ) ) {
+					$field_id = 0;
+					$content  = '';
+					/* Extract field id and html content */
+					foreach ( $form_data['html_id'] as $key => $value ) {
+						$field_id = $key;
```
jakejackson1: Won't this loop through the entire array and only set the last array item's `$key` and `$value` as `$field_id` and `$content` respectively? 

It feels like you should rename `$key` to `$field_id` and `$value` to `$content`, and then remove lines 121-123 inside this array. You'd then be able to remove lines 113-114.

jakejackson1 (changes requested):
Missing unit tests

jakejackson1:
The PR title and description are not accurate for this. Please correct.

This also needs a unit test before merging.

jestonihpi:
Hi @jakejackson1, I have added the missing unit test, kindly review this PR please. Thanks!

tests/phpunit/unit-tests/Helper/Fields/Test_Field_Repeater.php
```diff
@@ -80,4 +80,20 @@ public function test_repeater_add_entry_meta_method() {
...
+		$data          = $this->pdf_field->get_repeater_form_data( $this->pdf_field->form_data(), $this->pdf_field->value(), new \GF_Field_Repeater( $this->form['fields'][1] ) );
+		$repeater_data = $data['repeater']['999.Team'][0];
+
+		/* Verify that method correctly adds the HTML field with each corresponding indexes. */
+		$this->assertArrayHasKey( 17, $repeater_data );
+		$this->assertArrayHasKey( 'HTML', $repeater_data );
+		$this->assertArrayHasKey( 'HTML.17', $repeater_data );
+
+		/* Check that if HTML doesn't exist, method should not add HTML in form data. */
+		$this->assertNotTrue( isset( $repeater_data[99][17] ) );
```
jakejackson1: Use `assertArrayHasNotKey` instead.

tests/phpunit/unit-tests/Helper/Fields/Test_Field_Repeater.php
```diff
@@ -80,4 +80,20 @@ public function test_repeater_add_entry_meta_method() {
...
 		$this->assertSame( 'USD', $values[10] );
 	}
+
+	public function test_get_repeater_form_data() {
+		/* This method expects the third parameter as an GF_Field_Repeater Instance. */
+		$data          = $this->pdf_field->get_repeater_form_data( $this->pdf_field->form_data(), $this->pdf_field->value(), new \GF_Field_Repeater( $this->form['fields'][1] ) );
+		$repeater_data = $data['repeater']['999.Team'][0];
+
+		/* Verify that method correctly adds the HTML field with each corresponding indexes. */
+		$this->assertArrayHasKey( 17, $repeater_data );
```
jakejackson1, line 90: You should check the content of each of these three array keys to verify it is what you expect it to be.

tests/phpunit/unit-tests/Helper/Fields/Test_Field_Repeater.php
jakejackson1: I'm not clear how checking `$repeater_data[99][17]` verifies that no HTML fields are added to the sub-key. If no HTML field exists in the sub-repeater field in the form, and that sub-key only has the fields 200, 201, and 88 registered, why would checking the hardcoded field ID 17 on the sub key validate this behaviour?

jakejackson1 (changes requested):
@jestonihpi left you some notes

jestonihpi:
Hi @jakejackson1 , this PR was updated. I have removed the unnecessary test for checking if HTML is not added on non HTML fields. Since it requires an HTML Field to initialize the IF statement I've created, thus making it unnecessary.
Thank you!

src/Helper/Fields/Field_Repeater.php
```diff
@@ -107,6 +107,20 @@ public function get_repeater_form_data( $data, $value, $field ) {
...
+					$html = [];
+					/* Extract field id and html content */
+					foreach ( $form_data['html_id'] as $field_id => $content ) {
+						/* Format data, with the exact same order ( label.id,id,label ) with the other fields */
+						$html[ $class->field->label . '.' . $field_id ] = $content;
+						$html[ $field_id ]                              = $content;
+						$html[ $class->field->label ]                   = $content;
+					}
+
+					$data = array_replace_recursive( $data, [ $id => $html ] );
```
jakejackson1, line 122: @jestonihpi out of interest, what's the issue with assigning the HTML array keys/value directly to the `$data` array?

jakejackson1 (commented):
This looks good @jestonihpi. Just had a question about merging in the HTML info to the `$data` array.

src/Helper/Fields/Field_Repeater.php
jestonihpi: To make it look like a standard entry array @jakejackson1 iirc. 
If we put it directly into data it would look something like this :
image
Where it should be like this :
image

PR #1344, opened 2022-06-01
Updates on URL arguments 

jestonihpi:
## Description
Add an `escape` function for those url that need to be printed as html

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/Helper/Helper_PDF_List_Table.php
+++ b/src/Helper/Helper_PDF_List_Table.php
@@ -360,7 +360,7 @@ class Helper_PDF_List_Table extends WP_List_Table {
 		$delete_nonce    = wp_create_nonce( "gfpdf_delete_nonce_{$form_id}_{$item['id']}" );
 
 		$actions = [
-			'edit'      => '<a title="' . esc_attr__( 'Edit this PDF', 'gravity-forms-pdf-extended' ) . '" href="' . $edit_url . '">' . esc_html__( 'Edit', 'gravity-forms-pdf-extended' ) . '</a>',
+			'edit'      => '<a title="' . esc_attr__( 'Edit this PDF', 'gravity-forms-pdf-extended' ) . '" href="' . esc_url( $edit_url ) . '">' . esc_html__( 'Edit', 'gravity-forms-pdf-extended' ) . '</a>',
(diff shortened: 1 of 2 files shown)


jestonihpi:
Let me know if the PR for the ubuntu fix  is merged @jakejackson1 , so that I can reset this branch and finalize this PR.

src/Model/Model_Form_Settings.php
```diff
@@ -184,7 +184,7 @@ public function process_list_view( $form_id ) {
 		$controller->view->list(
 			[
 				'title'       => $this->data->title,
-				'add_new_url' => $add_new_url = add_query_arg( [ 'pid' => 0 ] ),
+				'add_new_url' => $add_new_url = esc_url( add_query_arg( [ 'pid' => 0 ] ) ),
```
jakejackson1: @jestonihpi this would be better handled at the point the URL is output in the HTML.

src/Helper/Helper_PDF_List_Table.php
```diff
@@ -353,7 +353,7 @@ public function column_template( $item ) {
 	 * @since 4.0
 	 */
 	public function column_name( $item ) {
-		$edit_url        = add_query_arg( [ 'pid' => $item['id'] ] );
+		$edit_url        = esc_url( add_query_arg( [ 'pid' => $item['id'] ] ) );
```
jakejackson1: @jestonihpi this would be better handled at the point the URL is output in the HTML.

jakejackson1 (changes requested):

src/View/html/FormSettings/list.php
```diff
@@ -34,7 +34,7 @@
 				<div class="tablenav top">
 					<div class="alignleft actions bulkactions"></div>
 					<div class="alignright">
-						<a class="button" href="<?= esc_url( $args['add_new_url'] ) ?>" aria-label="<?= esc_attr__( 'Add new PDF', 'gravity-forms-pdf-extended' ) ?>"><?php esc_html_e( 'Add New', 'gravityforms' ); ?></a>
+						<a class="button" href="<?= esc_url( esc_url( $args['add_new_url'] ) ) ?>" aria-label="<?= esc_attr__( 'Add new PDF', 'gravity-forms-pdf-extended' ) ?>"><?php esc_html_e( 'Add New', 'gravityforms' ); ?></a>
```
jakejackson1: @jestonihpi you've wrapped the output in `esc_url()` twice. One of these will need to be removed.

jakejackson1 (changes requested):

PR #1346, opened 2022-06-06
Don't delete top-level directory during cache cleanup

jakejackson1:
## Description

This PR increases performance by running two less I/O operations, and may also fix a race condition between the folder being deleted and getting recreated.

TODO:

- [x] Write unit test to ensure top-level directory is not deleted
- [x] Replicate this bug by writing code that will trigger this race condition. This should be doable by generating a PDF while the PDF cache is flushed repeatedly.

## Checklist:
- [x] I've tested the code.


jakejackson1:
Notify HS customers with the tag `race-condition-tmp-dir` once this issue has been fully investigated.

PR #1352, opened 2022-06-21
PDF Form Settings Permissions Change

### LINKED ISSUE #1342: Cannot enable user/editor access to a form's PDF settings without exposing GF's plugin settings
#### Description
As far as I can tell there's no way to allow certain users to access/configure a form's PDF settings (**Dashboard > Form > Settings > PDF**) without giving them the `gravityforms_edit_settings` cap, which opens up the Gravity Forms plugin configuration under **Dashboard > Gravity Forms > Settings**.

In contrast, the default GF form settings (e.g. **Form Settings**, **Confirmations**, **Notifications**) only require the `gravityforms_edit_forms` cap. This allows website editors and other users/roles with limited permissions to access a form's settings pages without also allowing them to mess with the GF plugin configuration. 

#### Step To Reproduce
N/A

#### Expected behavior
We should be able to allow access to a form's PDF settings pages without exposing plugin settings pages.

Assuming you want to maintain backwards compatibility here and aren't willing to switch `gravityforms_edit_settings` for `gravityforms_edit_forms`, maybe you can provide a filter that we can use to override the cap check? 

#### Screenshots
N/A

#### Additional context
N/A

jakejackson1:
## Description

Use correct permission when editing PDF Form Settings

Swaps `gravityforms_edit_settings` to `gravityforms_edit_forms` capability, brining it in line with the correct Gravity Forms capability configuration.

Also changes the permission for the font manager from `gravityforms_view_entries` to `gravityforms_edit_forms`.

Resolves #1342

## Testing instructions
1. Install User Role Manager Plugin
2. Create new role and only enable the `gravityforms_edit_forms` capability
3. Create a new user and give it this role
4. Install the User Switching plugin  
5. Switch to this new user and verify you can now add/edit/delete PDFs configured on Gravity Forms

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -114,6 +114,8 @@ If you aren't able to meet the v6 minimum requirements [you can download v5 whic
 * Security (Hardening): Escape any and all strings from the Gravity Forms form object, in any context (including PDFs)
 * Security (Hardening): The ?html=1 and ?data=1 developer helper parameters now only work in non-production environments, or when Gravity PDF Debug Mode is explicitly enabled
 * Security (Hardening): Prevent directory traversal when loading the various Gravity PDF UI components
+* Security (Hardening): Change PDF Form Settings capability check from `gravityforms_edit_settings` to `gravityforms_edit_forms`
+* Security (Hardening): Change Font Manager CRUD capability check from `gravityforms_view_entries` to `gravityforms_edit_forms`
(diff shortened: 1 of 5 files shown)


jestonihpi (approved):
Working as expected. Tested through  couple of user accounts.

PR #1353, opened 2022-06-22
Filter Notification messages exclusively to Gravity PDF pages.

### LINKED ISSUE #1345: After  new PDF form submission, message should show along side other add-on messages.
**Description**

**Step To Reproduce**
1. Go to Install an add-on that requires other dependencies .
2. Click on 'Form -> Settings->PDF->Add new'
3. Submit a new PDF template

**Expected behavior**
Success message should show along side with other  alerts.
image

**Screenshots**
image

**Additional context**
I believe we already did some update regarding this. AFAIK, we prepend our message to `.gform-settings__content`.

jestonihpi:
Create a separate method that returns our `errors` and `notices`  messages exclusively to Gravity PDF pages.

NOTE: This may need some improvements on how to properly initialize the Helper_Notice class exclusively to Gravity PDF pages as well.
## Testing instructions
Install Gravity Forms Akismet Add-On. Its default error message should not display on PDF settings.

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/Helper/Helper_Notices.php
+++ b/src/Helper/Helper_Notices.php
@@ -28,7 +28,7 @@ class Helper_Notices implements Helper_Interface_Actions {
 	 *
 	 * @since 4.0
 	 */
-	private $notices = [];
+	protected $notices = [];
 
 	/**
 	 * Holds any errors that we've triggered
@@ -37,38 +37,53 @@ class Helper_Notices implements Helper_Interface_Actions {
 	 *
 	 * @since 4.0
 	 */
-	private $errors = [];
+	protected $errors = [];
 
 	/**
-	 * Initialise our class defaults
-	 *
-	 * @return void
 	 * @since 4.0
-	 *
 	 */
-	public function init() {
+	public function init(): void {
 		$this->add_actions();
 	}
 
 	/**
 	 * Apply any actions needed to implement notices
 	 *
-	 * @return void
 	 * @since 4.0
-	 *
 	 */
-	public function add_actions() {
+	public function add_actions(): void {
 		add_action( $this->get_notice_type(), [ $this, 'process' ] );
+		add_action( 'init', [ $this, 'maybe_remove_non_pdf_messages' ] );
+	}
+
+	/**
+	 * Override GF notices on Gravity PDF pages
+	 *
+	 * @since 6.5
+	 */
+	public function maybe_remove_non_pdf_messages(): void {
+		if ( ! \GPDFAPI::get_misc_class()->is_gfpdf_page() ) {
+			return;
+		}
+
+		/* Remove existing notice */
+		remove_action( $this->get_notice_type(), [ $this, 'process' ] );
+
+		/* Delete Gravity Forms notices */
+		add_action( 'gform_admin_messages', [ $this, 'reset_gravityforms_messages' ], 999 );
+		add_action( 'gform_admin_error_messages', [ $this, 'reset_gravityforms_messages' ], 999 );
+
+		/* Show Gravity PDF Notices */
+		add_action( 'gform_admin_messages', [ $this, 'set_gravitypdf_notices' ], 1000 );
+		add_action( 'gform_admin_error_messages', [ $this, 'set_gravitypdf_errors' ], 1000 );
 	}
 
 	/**
 	 * Determine which notice should be triggered
 	 *
-	 * @return string
-	 *
 	 * @since 4.0
 	 */
-	private function get_notice_type() {
+	protected function get_notice_type(): string {
 		if ( is_multisite() && is_network_admin() ) {
 			return 'network_admin_notices';
 		}
@@ -82,18 +97,9 @@ class Helper_Notices implements Helper_Interface_Actions {
 	 * @param string $notice The message to be queued
 	 * @param string $class  The class that should be included with the notice box
 	 *
-	 * @return void
-	 *
 	 * @since 4.0
 	 */
-	public function add_notice( $notice, $class = '' ) {
-
-		if ( GFForms::is_gravity_page() ) {
-			GFCommon::add_message( $notice );
-
-			return;
-		}
-
+	public function add_notice( string $notice, string $class = '' ): void {
 		if ( empty( $class ) ) {
 			$this->notices[] = $notice;
 		} else {
@@ -107,18 +113,9 @@ class Helper_Notices implements Helper_Interface_Actions {
 	 * @param string $error The error message that should be added
 	 * @param string $class Any class names that should apply to the error
 	 *
-	 * @internal param string $notice The message to be queued
-	 *
 	 * @since    4.0
 	 */
-	public function add_error( $error, $class = '' ) {
-
-		if ( GFForms::is_gravity_page() ) {
-			GFCommon::add_error_message( $error );
-
-			return;
-		}
-
+	public function add_error( string $error, string $class = '' ) {
 		if ( empty( $class ) ) {
 			$this->errors[] = $error;
 		} else {
@@ -129,31 +126,19 @@ class Helper_Notices implements Helper_Interface_Actions {
 	/**
 	 * Check if we currently have a notice
 	 *
-	 * @return boolean
-	 *
 	 * @since 4.0
 	 */
-	public function has_notice() {
-		if ( count( $this->notices ) > 0 ) {
-			return true;
-		}
-
-		return false;
+	public function has_notice(): bool {
+		return count( $this->notices ) > 0;
 	}
 
 	/**
 	 * Check if we currently have an error
 	 *
-	 * @return boolean
-	 *
 	 * @since 4.0
 	 */
-	public function has_error() {
-		if ( count( $this->errors ) > 0 ) {
-			return true;
-		}
-
-		return false;
+	public function has_error(): bool {
+		return count( $this->errors ) > 0;
 	}
 
 	/**
@@ -163,13 +148,12 @@ class Helper_Notices implements Helper_Interface_Actions {
 	 *
 	 * @since 4.0
 	 */
-	public function clear( $type = 'all' ) {
-
-		if ( 'errors' === $type || 'all' === $type ) {
+	public function clear( string $type = 'all' ): void {
+		if ( in_array( $type, [ 'all', 'errors' ], true ) ) {
 			$this->errors = [];
 		}
 
-		if ( 'notices' === $type || 'all' === $type ) {
+		if ( in_array( $type, [ 'all', 'notices' ], true ) ) {
 			$this->notices = [];
 		}
 	}
@@ -177,11 +161,9 @@ class Helper_Notices implements Helper_Interface_Actions {
 	/**
 	 * Process our admin notice and error messages
 	 *
-	 * @return void
-	 *
 	 * @since 4.0
 	 */
-	public function process() {
+	public function process(): void {
 		foreach ( $this->notices as $class => $notice ) {
 			$include_class = ( ! is_int( $class ) ) ? $class : '';
 			$this->html( $notice, 'updated ' . $include_class );
@@ -199,11 +181,9 @@ class Helper_Notices implements Helper_Interface_Actions {
 	 * @param string $text  The message to be displayed
 	 * @param string $class The class name (updated / error)
 	 *
-	 * @return void
-	 *
 	 * @since 4.0
 	 */
-	private function html( $text, $class = 'updated' ) {
+	protected function html( string $text, string $class = 'updated' ): void {
 		$allow_form_elements = static function( $tags ) {
 			$tags['input'] = [
 				'type'  => true,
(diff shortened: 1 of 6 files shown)


src/Helper/Helper_Notices.php
```diff
@@ -210,4 +212,32 @@ private function html( $text, $class = 'updated' ) {
...
+	 *
+	 * @param array $messages  The message to be displayed
+	 *
+	 * @return array $messages
+	 *
+	 * @since 6.3
+	 */
+	public function filter_messages( $messages ) {
+		global $gfpdf;
+		$misc = new Helper_Misc( $gfpdf->log, $gfpdf->gform, $gfpdf->data );
```
jakejackson1: `\GPDFAPI::get_misc_class()` would be a better option.

jakejackson1 (changes requested):
This isn't the best way to tackle this. 

I recommend removing L91-95 so that the Gravity PDF notices don't get mixed up with the Gravity Forms notices.

I'd then create an individual method for the hooks on L62 and L63 and check if it's a Gravity PDF page. If it is a GPDF page, return the queued Gravity PDF notice or error message.

jakejackson1:
Also, missing unit tests.

src/Helper/Helper_Notices.php
```diff
@@ -210,4 +203,37 @@ private function html( $text, $class = 'updated' ) {
 		</div>
 		<?php
 	}
+
+	/**
+	 * If the current page belongs to Gravity PDF, add these hooks.
+	 *
+	 * @since 6.3
+	 */
+	public function maybe_gravitypdf_page() {
```
jakejackson1: This is a poorly-named method. Perhaps `maybe_remove_non_pdf_notices`?

src/Helper/Helper_Notices.php
```diff
@@ -210,4 +203,37 @@ private function html( $text, $class = 'updated' ) {
 		</div>
 		<?php
 	}
+
+	/**
+	 * If the current page belongs to Gravity PDF, add these hooks.
+	 *
+	 * @since 6.3
```
jakejackson1: This should be 6.4

src/Helper/Helper_Notices.php
```diff
@@ -210,4 +203,37 @@ private function html( $text, $class = 'updated' ) {
...
+	public function maybe_gravitypdf_page() {
+		if ( \GPDFAPI::get_misc_class()->is_gfpdf_page() ) {
+			add_action( 'gform_admin_messages ', [ $this, 'remove_non_gravity_pdf_messages' ] );
+			add_action( 'gform_admin_error_messages', [ $this, 'remove_non_gravity_pdf_messages' ] );
+		}
+	}
+
+	/**
+	 * Remove all Non-Gravity PDF messages.
+	 * Due GFCommon::display_admin_message that only display one message at time.
```
jakejackson1: I have zero clue what this means. Add appropriate context.

src/Helper/Helper_Notices.php
```diff
@@ -210,4 +203,37 @@ private function html( $text, $class = 'updated' ) {
...
+	 *
+	 * @param array $messages  The message to be displayed
+	 *
+	 * @return array $messages
+	 *
+	 * @since 6.3
```
jakejackson1: 6.4

src/Helper/Helper_Notices.php
```diff
@@ -210,4 +203,37 @@ private function html( $text, $class = 'updated' ) {
...
 	}
+
+	/**
+	 * If the current page belongs to Gravity PDF, add these hooks.
+	 *
+	 * @since 6.3
+	 */
+	public function maybe_gravitypdf_page() {
+		if ( \GPDFAPI::get_misc_class()->is_gfpdf_page() ) {
+			add_action( 'gform_admin_messages ', [ $this, 'remove_non_gravity_pdf_messages' ] );
```
jakejackson1: Remove space after `_messages` and `'`

src/Helper/Helper_Notices.php
```diff
@@ -112,7 +106,6 @@ public function add_notice( $notice, $class = '' ) {
 	 * @since    4.0
 	 */
 	public function add_error( $error, $class = '' ) {
-
 		if ( GFForms::is_gravity_page() ) {
```
jakejackson1: We did we remove this check from the notice method and not the error as well?

src/Helper/Helper_Notices.php
```diff
@@ -210,4 +203,37 @@ private function html( $text, $class = 'updated' ) {
...
+	 * Due GFCommon::display_admin_message that only display one message at time.
+	 *
+	 * @param array $messages  The message to be displayed
+	 *
+	 * @return array $messages
+	 *
+	 * @since 6.3
+	 */
+	public function remove_non_gravity_pdf_messages( $messages ) {
+		foreach ( $messages as $i => $message ) {
```
jakejackson1: The idea is to disregard `$messages` on Gravity PDF-related pages and return `$this->notices` or `$this->errors` instead, thereby ensuring that non-Gravity PDF notices/errors won't show up on Gravity PDF pages.

jakejackson1 (changes requested):
The PR description and title should be updated appropriately.

src/Helper/Helper_Notices.php
```diff
@@ -225,4 +213,51 @@ private function html( $text, $class = 'updated' ) {
...
+	 * @return array $this->errors
+	 *
+	 * @since 6.4
+	 */
+	public function return_pdf_error_messages( array $messages ): array {
+		if ( $this->has_error() ) {
+			return $this->errors;
+		}
+
+		return [];
```
jakejackson1: Let's reverse this so that if there are no errors we return an empty array, but if there are errors we return the errors. It makes the code more readable by putting the negative condition inside the IF statement and the positive condition in the top-level flow. The same goes for the notices.

tests/phpunit/unit-tests/test-notices.php
```diff
@@ -159,4 +159,13 @@ public function test_html_notice() {
 
 		$this->assertNotFalse( strpos( $html, $form ) );
 	}
+
+	/* Test empty $this->errors and $this->notices */
+	public function test_gform_admin_messages_hooks() {
```
jakejackson1, line 167: I think we need to enhance the tests here to check specifically that the registered errors/notices are returned on Gravity PDF pages that run the `gform_admin_error_messages` or `gform_admin_messages` filters, but are NOT returned if the pages are not Gravity PDF ones.

src/Helper/Helper_Notices.php
```diff
@@ -225,4 +213,51 @@ private function html( $text, $class = 'updated' ) {
...
+	/**
+	 * Return Gravity PDF error messages and disregard the existing messages.
+	 *
+	 * @param array $messages The message to be displayed
+	 *
+	 * @return array $this->errors
+	 *
+	 * @since 6.4
+	 */
+	public function return_pdf_error_messages( array $messages ): array {
```
jakejackson1: A better name for this method would be `override_gravityforms_admin_error_messages()`

src/Helper/Helper_Notices.php
```diff
@@ -225,4 +213,51 @@ private function html( $text, $class = 'updated' ) {
...
+	/**
+	 * Return Gravity PDF notice messages and disregard the existing messages.
+	 *
+	 * @param array $messages The message to be displayed
+	 *
+	 * @return array $this->notices
+	 *
+	 * @since 6.4
+	 */
+	public function return_pdf_notice_messages( array $messages ): array {
```
jakejackson1: A better name for this method would be `override_gravityforms_admin_messages()`

tests/phpunit/unit-tests/test-notices.php
jakejackson1: You can use `apply_filters()` to trigger the correct filters, but you'll need to override the globals to mock Gravity PDF vs non-Gravity PDF pages.

src/Helper/Helper_Notices.php
```diff
@@ -225,4 +213,51 @@ private function html( $text, $class = 'updated' ) {
 
 		remove_filter( 'wp_kses_allowed_html', $allow_form_elements );
 	}
+
+	/**
+	 * If the current page belongs to Gravity PDF, add these hooks.
+	 *
+	 * @since 6.4
```
jakejackson1: As 6.4 is now released, this should be updated to 6.5

src/Helper/Helper_Notices.php
```diff
@@ -225,4 +213,51 @@ private function html( $text, $class = 'updated' ) {
...
+	}
+
+	/**
+	 * Return Gravity PDF error messages and disregard the existing messages.
+	 *
+	 * @param array $messages The message to be displayed
+	 *
+	 * @return array $this->errors
+	 *
+	 * @since 6.4
```
jakejackson1: As 6.4 is now released, this should be updated to 6.5

src/Helper/Helper_Notices.php
```diff
@@ -225,4 +213,51 @@ private function html( $text, $class = 'updated' ) {
...
+	}
+
+	/**
+	 * Return Gravity PDF notice messages and disregard the existing messages.
+	 *
+	 * @param array $messages The message to be displayed
+	 *
+	 * @return array $this->notices
+	 *
+	 * @since 6.4
```
jakejackson1: As 6.4 is now released, this should be updated to 6.5

jakejackson1 (changes requested):
@jestonihpi this is looking a lot better since last time I checked it out. The unit tests are lacking though and I'd like to see more tests to verify this functions as expected.

src/Helper/Helper_Notices.php
```diff
@@ -59,6 +59,20 @@ public function init() {
...
+	}
+
+	/**
+	 * If the current page belongs to Gravity PDF, add these hooks.
+	 * Remove these hooks if the current page does not belong to Gravity PDF.
+	 *
+	 * @since 6.5
+	 */
+	public function maybe_remove_non_pdf_messages() {
+		if ( \GPDFAPI::get_misc_class()->is_gfpdf_page() ) {
```
jakejackson1: Let's reverse this condition and return early if not a GF page. It makes the flow easier to follow as you can have the main logic below it.

jakejackson1 (changes requested):
Per the Codecov alert on L241 and L258, we also need tests that verify an empty array is returned when on GPDF pages.

jestonihpi:
Hi @jakejackson1 , can you check on this PR as well? Thanks!

src/Helper/Helper_Notices.php
```diff
@@ -225,4 +228,38 @@ private function html( $text, $class = 'updated' ) {
...
+	 *
+	 * @return array $this->notices
+	 *
+	 * @since 6.5
+	 */
+	public function override_gravityforms_admin_messages( $messages ): array {
+		if ( ! $this->has_notice() ) {
+			return [];
+		}
+		return $this->notices;
```
jakejackson1: @jestonihpi for consistency, add a new line before the return statement, like in the method above.

jakejackson1 (changes requested):
@jestonihpi just a minor change for consistency and then this should be ready to go I think.

jestonihpi:
Copy @jakejackson1. Thanks!

jestonihpi:
Hey @jakejackson1 , this PR is now ready for your review.

jakejackson1:
@jestonihpi I found problems with the notices being displayed when any Gravity PDF extensions are also activated on the site. This is because each extension creates a new Notices object on initialisation and they all add their own independent `maybe_remove_non_pdf_messages` hook (and subsequent override hooks). 

The exact flow when two Gravity PDF extensions are activated and the Core plugin saves the PDF settings to display the "Successful" message is:

1. Core Registers Notices class
1. Previewer Registers Notices class
1. Bulk Generator Registers Notices class
1. Core Registers Override Hooks
1. Previewer Registers Override Hooks
1. Bulk Generator Registers Override Hooks
1. Core Override Triggers and returns the "Success" message
1. Previewer Override Triggers and returns empty array
1. Bulk Generator Override Triggers and returns empty array 
1. Gravity Forms displays no notices

I think the best way to handle this is to split up the `gform_admin_messages` and `gform_admin_error_messages` actions we register so that we empty any existing messages at priority 999 (`return []` and add any Gravity PDF messages at priority 1000 `return array_merge( $messages, $this->notices )`.

This will ensure Gravity PDF Core and Extensions can still have unique Notices classes, non-Gravity PDF notices are removed on Gravity PDF pages, and each plugin won't override each other's messages. 

Does that make sense?

jestonihpi:
None for now @jakejackson1 , but I will keep you posted for further questions.

jestonihpi:
Hi @jakejackson1 , I have updated the unit tests. I have separated the non-filtered test and filtered. So with this, non-filtered should only reuse the notice object in the init, while the filtered test needs to re initialize starting from init.

jestonihpi:
@jakejackson1 , converting this PR to draft for now. Its not working as expected. 
image

jakejackson1:
@jestonihpi I reverted your last change. I believe the Core Font installer / registered actions inside of `Controller_Actions` / certain folder permissions make use of the notice class to output messages to the end user. These notices can be shown outside Gravity PDF pages and this handles that use-case. 

I also committed a tweak that should handle the GF-messages registered outside the `gform_admin_messages` processing (needs further testing).

jestonihpi:
Copy @jakejackson1.

PR #1356, opened 2022-06-27
Harden Signed URL Secret Key Creation

jakejackson1:
## Description

Replace usage of wp_generate_password() which includes the `random_password` filter, to sodium_crypto_secretbox_keygen() with fallback to modified version of wp_generate_password() which doesn't include the above filter.

## Testing instructions

1. On new installation test the signed PDF URL feature as a logged out user with the [Restrict Owner Access setting enabled
](https://docs.gravitypdf.com/v6/users/setup-pdf#restrict-owner)
2. On an existing installation test the signed PDF URL feature as a logged out user with the [Restrict Owner Access setting enabled
](https://docs.gravitypdf.com/v6/users/setup-pdf#restrict-owner)

In both cases there should be no issues with the user viewing the PDF.

## Checklist:
- [x] I've tested the code.


jestonihpi (approved):
Working as expected, including invalid key

PR #1358, opened 2022-06-30
Better Sanitizing and Escaping of PDF URLs

jakejackson1:
## Description

The focus of this change is in the middleware and the PDF URL builder method. These changes are bundled together because the PDF URL generator builds signed PDF URLs, while the middleware verifies those URLs.

Additional log messages have been added to all PDF authentication middleware, or existing logs modified, to make it clearer what is occurring.

## Testing instructions

The same tests should be performed with both Pretty Permalinks enabled and disabled (Settings -> Permalinks).

1. Setup a form with a PDF configured 
2. On the Confirmation page add the [gravitypdf] shortcode and [all variants of it using all available attributes  ](https://docs.gravitypdf.com/v6/users/shortcodes-and-mergetags#building-the-shortcode)
3. Also on the Cnformation page, add the Gravity PDF merge tag [and all variants of it](https://docs.gravitypdf.com/v6/users/shortcodes-and-mergetags#building-the-merge-tag), both inside and outside a HTML anchor tag 
4. Submit the form and check that all links are valid, and all RAW URLs are valid when copying/pasting into the address bar 
5. When testing the signed URL, add a breakpoint at line 346 of Model_PDF.php and enable Xdebug. Attempt to view a signed PDF URL and ensure the debugger stops at the breakpoint each time.

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/Model/Model_Mergetags.php
+++ b/src/Model/Model_Mergetags.php
@@ -210,7 +210,7 @@ class Model_Mergetags extends Helper_Abstract_Model {
 
 				/* Everything is valid so get the URL and display */
 				$modifiers = explode( ':', $tag[2] ?? '' );
-				$url       = $this->pdf->get_pdf_url( $tag[1], $entry['id'], (bool) in_array( 'download', $modifiers, true ), (bool) in_array( 'print', $modifiers, true ), $url_encode );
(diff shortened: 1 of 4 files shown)


jestonihpi (approved):
Working as expected, verify unit tests.

PR #1380, opened 2022-07-19
Fix Field Select not displaying when the saved value is 0

### LINKED ISSUE #1377: Select Field with Zero Value Not Displaying in Core PDF
**Description**
If a Select field is added to the form with the "Show Values" option enabled and then an option is given the value of `0` (zero), that entire field will not get displayed in the PDF when that option is selected (provided the "Show Empty Fields" PDF option is disabled).

**Step To Reproduce**
1. Import this form 
[sample.json.txt](https://github.com/GravityPDF/gravity-pdf/files/9089616/sample.json.txt) (remove the .txt extension first)
2. Fill out the form and select the first option
3. View the PDF
4. The field will now get displayed

**Expected behavior**
Any Select option with a non-empty value (i.e `$value !== ''`) should get displayed

**Additional context**
In `Field_Select` the `empty()` method uses `empty( $value['value'] )` to check if the variable has a value. As `0` is a falsey value this returns `true` in this circumstance. Change the check to `$value['value'] === ''` instead.

jestonihpi:
## Description
This PR will check if the select value is "real" empty by update the `is_empty` conditon to `===` instead of `empty()` which returns false if the value is 0.

#1377

## Testing instructions
Create a Select field with one of the option value is 0.  When saved and viewed it should NOT display in the pdf.

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/tests/phpunit/unit-tests/Helper/Fields/Test_Field_Select.php
+++ b/tests/phpunit/unit-tests/Helper/Fields/Test_Field_Select.php
@@ -57,6 +57,27 @@ class Test_Field_Select extends WP_UnitTestCase {
 		remove_filter( 'gfpdf_field_is_empty_value_instead_of_label', '__return_false' );
 	}
 
+	public function test_is_empty_true() {
+
+		$pdf_field = new Field_Select( $this->gf_field, [ 'form_id' => $this->form['id'], 3 => '', ], \GPDFAPI::get_form_class(), \GPDFAPI::get_misc_class() );
+		$this->assertTrue( $pdf_field->is_empty() );
+
+		$pdf_field = new Field_Select( $this->gf_field, [ 'form_id' => $this->form['id'], 3 => null, ], \GPDFAPI::get_form_class(), \GPDFAPI::get_misc_class() );
+		$this->assertTrue( $pdf_field->is_empty() );
+	}
+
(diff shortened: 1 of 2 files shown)


jakejackson1 (changes requested):
Needs a unit test and the PR description needs to be updated with appropriate details.

tests/phpunit/unit-tests/Helper/Fields/Test_Field_Select.php
```diff
@@ -57,6 +57,21 @@ public function test_is_empty() {
 		remove_filter( 'gfpdf_field_is_empty_value_instead_of_label', '__return_false' );
 	}
 
+	public function test_is_empty_falsey() {
```
jakejackson1: @jestonihpi you should also feed the test a real string like you would expect in a real form entry to verify `is_empty` isn't triggered incorrectly.

jakejackson1 (changes requested):
Just need an extra test to verify the change.

jestonihpi:
Hi @jakejackson1, this PR is now ready for review.

jakejackson1 (approved):
The code looks good. Approved. 

When I get a chance, I'll run a live test to confirm the functionality and then merge. Nice job.

PR #1385, opened 2022-07-25
V5 environment

jakejackson1:
## Description

Update automated testing environment

## Checklist:
- [x] I've tested the code.


jakejackson1:
The E2E tests don't work as they are for an older version of Gravity Forms and we aren't going to update them, but all other tests pass.

PR #1397, opened 2022-08-03
Fix URL Escape issue when using Image Path

jakejackson1:
## Description

Resolves escaping issue when using a Windows-based local path in images.

`esc_url()` will completely remove a Windows-based local path and the PDF will show an image-not-found icon.

## Testing instructions

On a Windows machine with a local development environment (not containered), add a signature field to your form, fill out the form, and view the PDF. Prior to this update the signature wouldn't show. After the update it does.

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/Helper/Fields/Field_Post_Image.php
+++ b/src/Helper/Fields/Field_Post_Image.php
@@ -65,7 +65,8 @@ class Field_Post_Image extends Helper_Abstract_Fields {
 		$html = '';
 		if ( ! empty( $value['url'] ) ) {
 			$html  = '<a href="' . esc_url( $value['url'] ) . '" target="_blank">';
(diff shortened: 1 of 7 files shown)


jakejackson1:
TODO: write unit tests for the Signature and Post fields (the Slim plugin is no longer developed).

jakejackson1:
@jestonihpi could you test this PR in a Window's dev environment (without Docker or any virtualisation)? Users running Windows servers reported problems displaying signature images in PDFs. This PR should fix it.

jestonihpi (approved):
Confirmed that prior to this update, Signature is not showing on my laragon server. Once this branch is checked out, the Signature is showing and working as expected.

PR #1418, opened 2022-10-11
Add Product quantity on Field_Product

### LINKED ISSUE #1417: Include Quantity Value in Individual Product Fields when no linked Quantity Field included
**Description**
When Products are ungrouped (either with code or the Core Booster add-on), the Product field does not ever include the Quantity selected by the user. This is problematic if a separate Quantity field isn't added to the form as the PDF will only show the selected product and price, but not quantity.

jestonihpi:
## Description
Added the product quantity to Non Grouped products. If the Product's quantity field is disabled, the quantity should not display on the pdf, same goes to Product with linked quantity field.

#1417 
## Testing instructions

- Create a form with and assigned it to any core templates.
- Create 3 product fields: ( Default, Disabled product quantity and a Product with a Quantity field linked into it)
- Make sure you have Core Booster installed and set the Product grouping to none.
- Test if Default will show the quantity and the other two should not show the quantities.

## Screenshots 
image

## Checklist:
- [x] I've tested the code.


src/Helper/Fields/Field_Product.php
```diff
@@ -62,7 +62,7 @@ public function html( $value = '', $label = true ) {
 			if ( in_array( $this->field->get_input_type(), [ 'radio', 'select' ], true ) ) {
 				$html .= Kses::parse( $value['name'] . ' - ' . $value['price'] );
 			} else {
-				$html .= esc_html( $value['price'] );
+				$html .= esc_html( $value['price'] ) . ' (' . $value['quantity'] . ' ' . __( 'Qty', 'gravity-forms-pdf-extended' ) . ') ';
```
jakejackson1: @jestonihpi couple things here:

1. The quantity should not be included in the output if there is a quantity field in the form that links to this product. Otherwise the PDF will display the quantity twice (one in the Product field and once in the Quantity field). 

2. The quantity should not be included in the output if the quantity has been disabled in the Product field

3. The new string you added needs to go _inside_ the `esc_html()` function.

jakejackson1 (changes requested):
@jestonihpi left you some notes.

src/Helper/Fields/Field_Product.php
jestonihpi: Copy  @jakejackson1 , thanks!

jestonihpi:
This PR is now ready for review @jakejackson1. Thanks!

jakejackson1:
@jestonisumande I've kind of co-opted this PR to better handle the display of individual Pricing fields in the PDF. The changes make it much clearer exactly how much an individual product (+ options) costs when ungrouped from the Product Table. 

[Individual Pricing.pdf](https://github.com/GravityPDF/gravity-pdf/files/10151163/Individual.Pricing.pdf)

Unit tests will need to be written to verify the accuracy of the output. But this could use another test of eyes and thorough testing against a range of forms + pricing fields to check the validity of the output.

jakejackson1:
@jestonihpi we should also include the discount amount in the PDF when displaying an individual Coupon Code `-£ 4.00 (COUPON)`, instead of only displaying the coupon name `COUPON`.

jestonihpi:
Copy @jakejackson1.

jestonihpi:
If you have the time @jakejackson1, can you review this PR as well? Thanks!

src/Helper/Fields/Field_Products.php
```diff
@@ -318,13 +318,15 @@ public function value() {
 
 				/* Store product in $form_array array */
 				$form_array['products'][ $id ] = [
```
jakejackson1, line 320: @jestonihpi revert this formatting change. It isn't relevant to your PR.

src/Helper/Fields/Field_Coupon.php
```diff
@@ -60,7 +60,7 @@ public function __construct( $field, $entry, Helper_Abstract_Form $gform, Helper
 	 * @since 4.0
 	 */
 	public function html( $value = '', $label = true ) {
-		$value = $this->value();
+		$value = $this->field->get_value_export( $this->entry );
```
jakejackson1, line 63: @jestonihpi how difficult would it be to use this format for the discount code: `Label (Code): -$1.50`

Currently it uses this format: `Label (Code: -$1.50)`, which looks weird.

jakejackson1 (changes requested):
@jestonihpi looking good.

src/Helper/Fields/Field_Products.php
jestonihpi: I have added these data on form_array Jake for the quantity display changes.

PR #1428, opened 2022-11-09
Allow a maximum of 8 User-Defined CSS Classes in Core Template

### LINKED ISSUE #1425: Only allow a maximum of 8 User-Defined CSS Classes in Core Templates
**Description**
[There is a significant memory problem in mPDF when a HTML tag has a lot of classes](https://github.com/mpdf/mpdf/issues/1753). To reduce the risk of this occurring in Gravity PDF we should [limit the `$field->cssClass` property to a maximum of 10 classes](https://github.com/GravityPDF/gravity-pdf/blob/development/src/Helper/Helper_Abstract_Fields.php#L348).

**Step To Reproduce**
1. Add 25 unique classes to the Custom CSS Class setting in a field in the Form Editor
2. Setup a Core template on the form
3. Attempt to view the PDF
4. Get memory exhausted error

**Expected behavior**
The PDF to generate without any errors

**Additional context**
Per the issue upstream in mPDF, there isn't a simple fix to this problem. The only workaround I can think of at the moment is to try limit the number of CSS classes that are added to a HTML tag.

jestonihpi:
This PR set a limited amount of custom CSS per field in PDF generation. We still process the PDF but were going to ignore the rest of these custom CSS and log the "excluded" along side with its field information.

WIP - Unit test
## Description

#1425 

## Testing instructions

In the Form field settings, add 8 or more user defined CSS class. 
Check the PDF's html (?html=1) to see if limits the classes.

## Screenshots 
image
image

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/Helper/Helper_Abstract_Fields.php
+++ b/src/Helper/Helper_Abstract_Fields.php
@@ -345,7 +345,7 @@ abstract class Helper_Abstract_Fields {
 		$label = esc_html( $this->get_label() );
 		$type  = $this->field->get_input_type();
 
-		$html = '<div id="' . esc_attr( 'field-' . $this->field->id ) . '" class="gfpdf-field ' . esc_attr( 'gfpdf-' . $type ) . ' ' . esc_attr( $this->field->cssClass ) . '">
+		$html = '<div id="' . esc_attr( 'field-' . $this->field->id ) . '" class="gfpdf-field ' . esc_attr( 'gfpdf-' . $type ) . ' ' . esc_attr( $this->get_field_classes() ) . '">
 					<div class="inner-container">';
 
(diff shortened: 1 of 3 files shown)


jestonihpi:
Hi @jakejackson1, I have added the unit test for this pr and updated the logic of the get_field_classes method.

jakejackson1:
@jestonihpi ah, I see the logger isn't available in this class directly. Because of the extra verbosity required to log messages here, I think we should just remove it and silently drop and additional classes after #8. That should condense this new method down to a line or two. 

Also, this would be for Gravity PDF 6.5.2 and the `@version` docblock should be updated accordingly.

PR #1430, opened 2022-11-11
Prevent Fatal error when processing Kses::output with null value.

### LINKED ISSUE #1427: Ensure any use of the `Kses` class methods always get passed a String
**Description**

[Got reports of a fatal error](https://wordpress.org/support/topic/error-when-trying-to-create-pdf/#post-16165506) due to a non-string getting passed to one of our `Kses` class methods, which requires a string. While I haven't been able to replicate, we should guard against this by verifying the $html we pass to this class is infact a string.

jestonihpi:
This PR is a safety precaution when running Kses::output on null values. Added an `is_string` check before the said output should fix this [problem ](https://wordpress.org/support/topic/error-when-trying-to-create-pdf/#post-16165506) posted in our wordpress.org support.

## Description

#1427 

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/Statics/Kses.php
+++ b/src/Statics/Kses.php
@@ -27,7 +27,7 @@ class Kses {
 	 *
 	 * @since 6.4.0
 	 */
-	public static function output( string $html ): void {
+	public static function output( $html ): void {
 		/* phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped */
 		echo self::parse( $html );
 	}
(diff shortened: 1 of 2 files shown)


jakejackson1:
@jestonihpi Upon thinking about this further, I think the best solution would be to edit the `Kses` class directly, remove the `string` type from the property and then check for a string at the beginning of the associated method(s). If it isn't a string, throw a `GravityPdfException` at this point so our error handlers can capture this.

jakejackson1:
@jestonihpi ah, I see the logger isn't available in this class directly. Because of the extra verbosity required to log messages here, I think we should just remove it and silently drop and additional classes after #8. That should condense this new method down to a line or two. 

Also, this would be for Gravity PDF 6.5.2 and the `@version` docblock should be updated accordingly.

jestonihpi:
Hi @jakejackson1, this PR is now ready for review.

src/Statics/Kses.php
```diff
@@ -25,9 +27,14 @@ class Kses {
 	 *
 	 * @return void
 	 *
+	 * @throws GravityPdfException
 	 * @since 6.4.0
 	 */
-	public static function output( string $html ): void {
+	public static function output( $html ): void {
+		/* If the $html is not a valid string, throw an exception if this returns false. */
+		if ( ! is_string( $html ) ) {
```
jakejackson1: @jestonihpi I think moving this into self::parse() would be better. It would cover any uses of the Kses class that way.

jestonihpi: Copy @jakejackson1.

PR #1449, opened 2023-02-13
Disable Queue Task when Gravity Forms Async Notifications is enabled.

### LINKED ISSUE #1444: Disable PDF Background Processing when Gravity Forms Async Notifications Enabled
**Description**

In Gravity Forms 2.6.9 they added support for Async Notifications: https://docs.gravityforms.com/gform_is_asynchronous_notifications_enabled/

If this is enabled, PDF Background Processing should be disabled:

```
$is_asynchronous = gf_apply_filters( array(
	'gform_is_asynchronous_notifications_enabled',
	$form_id,
), false, $event, $notifications_to_send, $form, $entry, [] );

if ( $is_asynchronous ) {
   // disable PDF Background Processing
}
```

It would be good to detect if this filter is enabled and remove the global PDF Background Processing setting from the UI too.

jestonihpi:
Prior to Gravity Form 2.6.9 release which introduced this new filter `gform_is_asynchronous_notifications_enabled`. We need to check its value when getting the queued task on form. If this is set to true we need to disable queued PDFs and notifications as well.

Missing Unit Test

## Description

Controller_Pdf_Queue class get_queuet_tasks method : get the value of the filter `gform_is_asynchronous_notifications_enabled` return $queue_data asap if its `true`

#1444 

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/Controller/Controller_Pdf_Queue.php
+++ b/src/Controller/Controller_Pdf_Queue.php
@@ -29,16 +29,7 @@ if ( ! defined( 'ABSPATH' ) ) {
  *
  * @since   5.0
  */
-class Controller_Pdf_Queue extends Helper_Abstract_Controller implements Helper_Interface_Actions, Helper_Interface_Filters {
-
-	/**
-	 * Holds the abstracted Gravity Forms API specific to Gravity PDF
-	 *
-	 * @var Helper_Form
-	 *
-	 * @since 5.0
-	 */
-	protected $gform;
+class Controller_Pdf_Queue extends Helper_Abstract_Controller {
 
 	/**
 	 * Holds our log class
@@ -64,13 +55,11 @@ class Controller_Pdf_Queue extends Helper_Abstract_Controller implements Helper_
 	protected $queue;
 
 	/**
-	 * Determines if our PDF queue should execute
+	 * @var array An array containing the notification objects to be sent using background processing during a request
 	 *
-	 * @var bool
-	 *
-	 * @since 5.0
+	 * @since 6.11.0
 	 */
-	protected $disable_queue = false;
+	protected $form_async_notifications = [];
 
 	/**
 	 * Set up our dependencies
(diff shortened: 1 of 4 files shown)


src/Controller/Controller_Pdf_Queue.php
```diff
@@ -278,6 +279,27 @@ protected function get_queue_tasks( $entry, $form, $notifications = [] ) {
...
+			false,
+			$notifications['event'],
+			$notifications['id'],
+			$form,
+			$entry,
+			[]
+		);
+
+		if ( $is_asynchronous ) {
+			return $queue_data;
```
jakejackson1: @jestonihpi looks like a good spot for it, considering we need to pass in the notification ID and event type being processed.

Couple things of note:

* `$queue_data` may not be empty if the `gfpdf_queue_initialise` filter is used and returns a value
* `$notifications['event']` is produce a PHP notice because the array key doesn't exist. `$notifications` is an array which, when looped over, will contain an individual notification config that includes the `event` key. Looking at the GF codebase, you should be able to take the event from the first notification configuration in this array. 
* `$notifications_to_send` should be an array of all Notification IDs that are sent, not a single ID. Like the above, this will produce a PHP notice. Use array_column to get the ID from all notifications.

src/Controller/Controller_Pdf_Queue.php
```diff
@@ -210,6 +210,7 @@ public function do_we_disable_notification( $default, $notification, $form, $ent
 	 * @since 5.0
 	 */
 	public function queue_async_form_submission_tasks( $entry, $form ) {
+
```
jakejackson1, line 208: Remove this extra line

src/Controller/Controller_Pdf_Queue.php
```diff
@@ -278,6 +279,27 @@ protected function get_queue_tasks( $entry, $form, $notifications = [] ) {
 
 		$queue_data = apply_filters( 'gfpdf_queue_initialise', [], $entry, $form );
 
+		/**
+		 * Check if Asynchronous notification is enabled for this form.
+		 * https://docs.gravityforms.com/gform_is_asynchronous_notifications_enabled/
+		 */
+		$is_asynchronous = gf_apply_filters(
```
jakejackson1: Let's move this to another method in this class and then check if `! $is_asyncronous` on line 214.

jakejackson1 (changes requested):
We probably need to display `queue_async_resend_notification_tasks()` if async is enabled too, right? Or does Gravity Forms not do async notifications when resending?

src/Controller/Controller_Pdf_Queue.php
```diff
@@ -278,6 +279,27 @@ protected function get_queue_tasks( $entry, $form, $notifications = [] ) {
...
+			[]
+		);
+
+		if ( $is_asynchronous ) {
+			return $queue_data;
+		}
+
 		/* Queue up the PDF generation callback */
 		if ( count( $pdfs ) > 0 ) {
 			$notifications = ( count( $notifications ) > 0 ) ? $notifications : $this->get_active_notifications( $form, $entry );
```
jakejackson1: Move to here: `array_filter( $notifications, function( $notifications ) { return ! apply_filters( 'gform_is_asynchronous_notifications_enabled' ... ) }`

jestonihpi:
@jakejackson1 , can you take peek at this PR when you're available. I tried working on the approach you've mentioned on the initial meeting. I was unable work my way around it. 

Not sure how to properly set the unit test for this 🤔

PR #1453, opened 2023-03-22
Fix Consent Field

### LINKED ISSUE #1450: Consent field showing up in Core PDF when not checked
**Description**
If a Consent field has a value in the "Description" setting, it will always be considered _not empty_ by Gravity PDF, even if a user didn't check the consent checkbox. This is contrary to how Gravity Forms handles consent fields in the Entry Details page, which always treats unticked Consent fields as empty.

**Step To Reproduce**
1. Add a Consent field to your form and include a value in the "Description" setting.
2. Add a Name field to your form and then save.
3. Setup a Core PDF on the form and ensure the _Show Empty Fields_ option is disabled
4. Submit the form and leave the Consent field unchecked
4. View the PDF

**Expected behavior**
The Consent field would not be shown in the PDF. Currently, it is showing.

**Screenshots**
Screenshot 2023-02-17 at 9 23 16 am

jestonihpi:
## Description
If the consent field is empty, it should not display on PDF, just like GravityForms handling empty consent field. We have added a new filter that allows users to bypass it manually.

Note: Missing Unit tests

#1450 
## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/Helper/Fields/Field_Consent.php
+++ b/src/Helper/Fields/Field_Consent.php
@@ -61,12 +61,17 @@ class Field_Consent extends Helper_Abstract_Fields {
 	 * @since 5.1
 	 */
 	public function html( $value = '', $label = true ) {
-		$value = $this->value();
+		$value       = $this->value();
+		$has_consent = $value['value'];
 
-		$has_consent = ! empty( $value['value'] );
+		/* Allow people to remove the consent field if a user didn't consent */
+		if ( ! $has_consent && apply_filters( 'gfpdf_hide_consent_field_if_empty', false, $this->field, $this->entry, $this->form ) ) {
+			return '';
+		}
 
-		$markup      = $has_consent ? $this->get_consented_markup( $value['label'] ) : $this->get_non_consent_markup();
-		$description = strlen( $value['description'] ) > 0 ? sprintf( '<div class="consent-text">%s</div>', $value['description'] ) : '';
+		$markup = $has_consent ? $this->get_consented_markup( $value['label'] ) : $this->get_non_consent_markup();
+
+		$description = ! empty( $value['description'] ) ? sprintf( '<div class="consent-text">%s</div>', $value['description'] ) : '';
 
 		/* consent has been given, determine order of description */
 		if ( empty( $value['placement'] ) || $value['placement'] === 'below' ) {
--- a/tests/phpunit/unit-tests/Helper/Fields/Test_Field_Consent.php
+++ b/tests/phpunit/unit-tests/Helper/Fields/Test_Field_Consent.php
@@ -1,0 +1,96 @@
+<?php
+
+declare( strict_types=1 );
+
+namespace GFPDF\Helper\Fields;
+
+use WP_UnitTestCase;
+
+/**
+ * @package     Gravity PDF
+ * @copyright   Copyright (c) 2023, Blue Liquid Designs
+ * @license     http://opensource.org/licenses/gpl-2.0.php GNU Public License
+ */
+
+/**
+ * @group   helper
+ * @group   fields
+ */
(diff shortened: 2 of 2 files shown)


src/Helper/Fields/Field_Consent.php
```diff
@@ -143,4 +139,20 @@ protected function get_non_consent_markup() {
...
+
+	/**
+	 * Add new filter gfpdf_hide_consent_field_if_empty that allows users to manually show consent field if empty on the pdf.
+	 *
+	 * @return string
+	 *
+	 * @since 6.5.4
+	 */
+	public function is_empty() {
+		if ( apply_filters( 'gfpdf_hide_consent_field_if_empty', true, $this->field, $this->entry, $this->form ) ) {
```
jakejackson1: This should be `false` by default.

src/Helper/Fields/Field_Consent.php
```diff
@@ -143,4 +139,20 @@ protected function get_non_consent_markup() {
...
+			$value = $this->value();
+			return empty( $value['value'] ); /* hide if empty */
+		}
+
+		return '';
```
jakejackson1: This should return `false`

src/Helper/Fields/Field_Consent.php
```diff
@@ -143,4 +139,20 @@ protected function get_non_consent_markup() {
 			__( 'Consent not given.', 'gravity-forms-pdf-extended' )
 		);
 	}
+
+	/**
+	 * Add new filter gfpdf_hide_consent_field_if_empty that allows users to manually show consent field if empty on the pdf.
+	 *
+	 * @return string
```
jakejackson1: `@return bool`

src/Helper/Fields/Field_Consent.php
```diff
@@ -143,4 +139,20 @@ protected function get_non_consent_markup() {
 			__( 'Consent not given.', 'gravity-forms-pdf-extended' )
 		);
 	}
+
+	/**
+	 * Add new filter gfpdf_hide_consent_field_if_empty that allows users to manually show consent field if empty on the pdf.
```
jakejackson1: Change to "Check if the field is empty"

src/Helper/Fields/Field_Consent.php
```diff
@@ -61,17 +61,13 @@ public function __construct( $field, $entry, Helper_Abstract_Form $gform, Helper
...
-		/* Return empty string if consent field is unticked. */
-		if ( empty( $value['value'] ) ) {
+		$value       = $this->value();
+		$has_consent = $value['value'];
+		if ( $this->is_empty() ) {
 			return '';
 		}
 
-		$markup = $has_consent ? $this->get_consented_markup( $value['label'] ): $this->get_non_consent_markup();
+		$markup      = $has_consent ? $this->get_consented_markup( $value['label'] ) : $this->get_non_consent_markup();
```
jakejackson1: Remove these changes

src/Helper/Fields/Field_Consent.php
```diff
@@ -61,17 +61,13 @@ public function __construct( $field, $entry, Helper_Abstract_Form $gform, Helper
...
-		global $gfpdf;
-		$value = $this->value();
-		$has_consent = false;
-		$input = $gfpdf;
-
-		/* Return empty string if consent field is unticked. */
-		if ( empty( $value['value'] ) ) {
+		$value       = $this->value();
+		$has_consent = $value['value'];
+		if ( $this->is_empty() ) {
```
jakejackson1: I don't think you need to include these changes here. The field should be filtered out automatically at https://github.com/GravityPDF/gravity-pdf/blob/development/src/View/View_PDF.php#L484

jakejackson1 (changes requested):
@jestonihpi feedback added

PR #1485, opened 2023-12-07
Allow some HTML in product / option / shipping field labels

jakejackson1:
## Description

Updates the Product table to allow some HTML in the field and choices labels. 

For legacy reasons, $form_data auto-escapes HTML. To display HTML in field labels we need to unescape the info first, and then run it through wp_kses_post(). 

## Testing instructions
1. Add Product, Option, and Shipping fields to a form
2. Include HTML in the label and option values
3. Submit the form
4. View a Core PDF
5. Ensure the text is formatted correctly

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/src/Helper/Fields/Field_Products.php
+++ b/src/Helper/Fields/Field_Products.php
@@ -153,7 +153,7 @@ class Field_Products extends Helper_Abstract_Fields {
 							<tr>
 								<td>
 									<div class="product_name">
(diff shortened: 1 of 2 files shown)


jakejackson1:
Notify https://secure.helpscout.net/conversation/2445795287/25696/ upon release

PR #1508, opened 2024-03-21
Fix PHP notice when $rule is not a valid array

jakejackson1:
## Description

Return early when $rule is not an array.

Had use report of the following PHP warning showing multiple times in their PDF document:

```
Warning: Trying to access array offset on value of type null in /wp-content/plugins/gravity-forms-pdf-extended/src/Controller/Controller_Form_Settings.php on line 346
```

## Testing instructions
The only way I could replicate this was by using a filter to modify the rule:

```
add_filter( 'gform_rule_pre_evaluation', function ( $rule ) {
	return null;
} );
```

The PR fixes the Gravity PDF-related errors. But if you use the filter, it does cause other errors to be triggered in the PDF by Gravity Forms. It's a case of bad input/bad output, so fixing the input problem (removing the bad filter) is the way forward.

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -109,6 +109,7 @@ Gravity PDF can be run on most modern shared web hosting without any issues. It 
 
 = 6.9.1 =
 * Security: Disable the Signed URL feature in the [gravitypdf] shortcode when a URL parameter provides the entry ID (e.g. Page Confirmations)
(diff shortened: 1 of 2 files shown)


jakejackson1:
TODO: replicate the issue

Issue #1527, opened 2024-04-30
Improve Template Caching


jakejackson1:
**Description**

The additional caching done on installed templates is causing some friction for some developers. Specifically those that install templates via FTP or a File Manager. 

1. Research viability of generating a "cache" file that includes a JSON dump of all template information. Clearing the cache is as simple as deleting the file. 
2. Research viability of disabling the caching when the WP environment is in development

jakejackson1:
Add unit tests for the Debug class when tackling this https://app.codecov.io/gh/GravityPDF/gravity-pdf/pull/1556/blob/src/Statics/Debug.php

jakejackson1:
Went with a different approach. In https://github.com/GravityPDF/gravity-pdf/pull/1641 the list of installed templates are only cached for the duration of the request. That will solve most issues being reported.

PR #1540, opened 2024-06-13
PDF Cache

### LINKED ISSUE #1539: Caching - add clear cache tool + dev mode bypass
## Problem

Currently, PDFs are saved to disk during form submission if the `gfpdf_post_save_pdf` hook is used, if the PDF is included as an email attachment, when using the Previewer add-on, or using the API.

Those PDFs are then cleaned up from disk:

* At the end of the form submission process
* At the end of the background processing queue
* After the Resend Notification process
* Every 12 hours (down from 24 hours)

If multiple notifications are sent on form submission with the same PDF attached, the PDF on disk will be used. 

When the new Gravity Forms Async Notification feature is used, the PDFs are not currently cleaned up from disk, but this could be done via the `gform_post_process_async_notifications` hook. 

Over the years various problems have cropped up due to the short lifespan of the temporary PDFs, or PDFs not being cleaned up at a suitable time. Some examples include:

* https://github.com/GravityPDF/gravity-pdf/issues/1403
* https://github.com/GravityPDF/gravity-pdf/issues/1370
* Async email sending e.g. WP Offload SES

As you can see the lifecycle of the PDF being saved to disk is currently unpredictable, and dynamically generating PDFs on demand over a short period of time is not performant. 

## Solution

Standardize the PDF lifecycle so it's more predicable and performant:

1. If a PDF is generated (viewed/downloaded/emailed), it will be saved to the temporary directory
1. If a PDF already exists in the temporary directory, the cached copy will be served
1. Remove all existing cleanup routines and replace it with a scheduled task that deletes PDFs older than 1 hour
1. ~If a form or entry is updated/deleted, the associated PDFs will be cleanup from the temporary directory~ The 1 hour cleanup routine will handle this automatically.

The short-lived disk cache strikes a balance between security, performance, and easy-of-use for first- and third-party plugins. If emails are delayed, the PDF attachment files will still be available for at least 1 hour to allow for (re)sending. If the same PDF is viewed many times in a short period of time, the cached version will be served. Provided stale copies of the PDF are automatically deleted from disk when form/entry changes are made, users should always receive up-to-date documents. 

Gravity PDF is making use of caching more and more throughout the codebase (templates, fonts, PDFs). Should the caches become stale, we need a simple solution for users to flush all caches. Currently this is tied to _Debug Mode_, but it's a toggle on/off feature – and not a one-off action – which confuses some users. 

To make this simpler for users:

1. Add a new Clear Cache tool that will handle all templates/fonts/pdfs cache clearing
2. Remove the UI for Debug Mode and auto-enable it when `WP_DEBUG` is enabled and/or a new filter is used

jakejackson1:
## Description

WIP.

PR to resolve #1539

Notes to self:

1. How to handle auto-cache purging when template is updated? Relevant https://github.com/GravityPDF/gravity-pdf/issues/1527

## Checklist:
- [ ] I've tested the code.

### DIFF
--- a/api.php
+++ b/api.php
@@ -352,48 +352,6 @@ final class GPDFAPI {
 	}
 
 	/**
-	 * Retrieve the filename of the PDF document
-	 * The .pdf extension is not returned by default (pass true to the 3rd param)
-	 *
-	 * See https://docs.gravitypdf.com/v6/developers/api/get_pdf_filename/ for more information about this method
-	 *
-	 * @param int    $entry_id          The Gravity Form entry ID
-	 * @param string $pdf_id            The Gravity PDF ID number (the pid number in the URL when viewing a setting in the admin area)
-	 * @param bool   $include_extension Whether to include the .pdf extension in the filename
-	 *
-	 * @return string|WP_Error   Return the filename of the PDF, or a WP_Error on failure
-	 *
-	 * @return string|WP_Error
-	 *
-	 * @since 6.12
-	 */
-	public static function get_pdf_filename( $entry_id, $pdf_id, bool $include_extension ) {
-		$form_class = static::get_form_class();
-
-		/* Get our entry */
-		$entry = $form_class->get_entry( $entry_id );
-		if ( is_wp_error( $entry ) ) {
-			return new WP_Error( 'invalid_entry', esc_html__( 'Make sure to pass in a valid Gravity Forms Entry ID', 'gravity-forms-pdf-extended' ) );
-		}
-
-		/* Get our settings */
-		$setting = static::get_pdf( $entry['form_id'], $pdf_id );
-		if ( is_wp_error( $setting ) ) {
-			return new WP_Error( 'invalid_pdf_setting', esc_html__( 'Could not located the PDF Settings. Ensure you pass in a valid PDF ID.', 'gravity-forms-pdf-extended' ) );
-		}
-
-		/** @var \GFPDF\Model\Model_PDF $pdf */
-		$pdf = static::get_mvc_class( 'Model_PDF' );
-
-		$filename = $pdf->get_pdf_name( $setting, $entry );
-		if ( $include_extension ) {
-			$filename .= '.pdf';
-		}
-
-		return $filename;
-	}
-
-	/**
 	 * Retrieve an array of the global Gravity PDF settings (this doesn't include individual form configuration details - see GPDFAPI::get_form_pdfs)
 	 *
 	 * See https://docs.gravitypdf.com/v6/developers/api/get_plugin_settings/ for more information about this method
--- a/api.php
+++ b/api.php
@@ -352,6 +352,48 @@ final class GPDFAPI {
 	}
 
 	/**
+	 * Retrieve the filename of the PDF document
+	 * The .pdf extension is not returned by default (pass true to the 3rd param)
+	 *
+	 * See https://docs.gravitypdf.com/v6/developers/api/get_pdf_filename/ for more information about this method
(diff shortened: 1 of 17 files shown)


src/Controller/Controller_Pdf_Queue.php
```diff
@@ -241,28 +236,6 @@ public function queue_async_tasks( $form, $entry ) {
...
-	 * Delete PDFs from disk once all tasks are processed
-	 *
-	 * @param array $form
-	 * @param array $entry
-	 *
-	 * @return void
-	 *
-	 * @since 6.11.0
-	 */
-	public function queue_cleanup_task( $form, $entry ) {
```
jakejackson1, line 259: @jakejackson1 re-add and deprecate.

src/Helper/Helper_PDF.php
```diff
@@ -874,13 +880,9 @@ protected function load_html( $args = [] ) {
 	 */
 	protected function maybe_display_raw_html( $html ) {
 
+		/* @TODO - move to shared file */
```
jakejackson1, line 883: @jakejackson1 make you use `rgget( 'html' ) && Debug::is_enabled_and_can_view()`

src/Model/Model_PDF.php
```diff
@@ -231,17 +240,98 @@ public function process_pdf( $pid, $lid, $action = 'view' ) {
 			return $middleware;
 		}
 
-		/* Add backwards compatibility support for certain settings */
-		$settings = $this->apply_backwards_compatibility_filters( $settings, $entry );
+		/*
+		 * Normalize the PDF action
+		 * The PDF cache introduced in 6.12 relies on a hash generated from the form, entry, and pdf settings
```
jakejackson1, line 245: @jakejackson1  Add more info about editing the form, entry or PDF settings auto-expires a PDF so the document is always fresh.

api.php
```diff
@@ -351,6 +351,48 @@ public static function delete_pdf( $form_id, $pdf_id ) {
...
+	 * @param string $pdf_id            The Gravity PDF ID number (the pid number in the URL when viewing a setting in the admin area)
+	 * @param bool   $include_extension Whether to include the .pdf extension in the filename
+	 *
+	 * @return string|WP_Error   Return the filename of the PDF, or a WP_Error on failure
+	 *
+	 * @return string|WP_Error
+	 *
+	 * @since 6.12
+	 */
+	public static function get_pdf_filename( $entry_id, $pdf_id, bool $include_extension ) {
```
jakejackson1: @jakejackson1 move to another PR. Not required here.

Will also add an API for getting the field markup for a PDF.

src/Statics/Cache.php
```diff
@@ -0,0 +1,51 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * @since 6.12.0
+ */
+class Cache {
+
+	protected static $template_tmp_location = null;
```
jakejackson1, line 27: @jakejackson1 add docblocks

src/Statics/Debug.php
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+/* Exit if accessed directly */
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * @since 6.12.0
+ * @TODO
```
jakejackson1: @jakejackson1 add docblocks

jakejackson1:
Superseded by https://github.com/GravityPDF/gravity-pdf/pull/1556

Issue #1403, opened 2022-08-15
Background Processing Conflict with WooCommerce GF Add-on


jakejackson1:
**Description**
When the background processing feature is enabled in Gravity PDF and a Notification is sent using the Woo GF add-on after checkout via the `GFAPI::send_notifications()` API, a critical error is shown in the Gravity PDF logs and the notification email isn't sent.

jestonihpi:
Hi @jakejackson1. I may need some help on how to replicate this : Is this the correct Add-on? 
![image](https://user-images.githubusercontent.com/434866/186299460-46191f2c-8c8c-4a0d-9786-e7ab92cfb000.png)
 I have installed woo and set up a demo product with no luck.

jestonihpi:
Updates @jakejackson1 . I think I have now properly set up my Woo addon. Another thing was I cant find any error on the logs :
![image](https://user-images.githubusercontent.com/434866/186312198-628ee7bf-45e0-43c1-ad0a-378b1ecaf325.png)

It seems I need to enable the notification though :
![image](https://user-images.githubusercontent.com/434866/186312291-4751caca-4c20-46d9-ac8c-0f24b56719f3.png)

jestonihpi:
Hi @jakejackson1. Unfortunately I was not able to replicate this issue on my end :
![image](https://user-images.githubusercontent.com/434866/186321626-e6d0c507-beb4-48dc-80dd-e9f59dca5f7d.png)

I may need some assistance on this ticket.

Thanks.

jakejackson1:
@jestonihpi sorry, I created this issue real quick and didn't fill in all the details.

[When linking the Woo Product to a Gravity Forms](https://woocommerce.com/document/woocommerce-gravity-forms-product-addons/#section-17), you will need to enable the "Update Payment and Transaction Details" and "Send Notifications" options.

On the form itself, configure a PDF that attaches to a configured Notification email.

Turn on the PDF Background Processing Feature.

Purchase the Woo Product with the GF Attached and complete the checkout process.

jestonihpi:
Thanks for the clarification @jakejackson1.

jestonihpi:
Hi 
[14003.log](https://github.com/GravityPDF/gravity-pdf/files/9486336/14003.log)
@jakejackson1 , can you verify this file if its the same log in your end?

Thanks.

jakejackson1:
Could not replicate in latest version

Issue #863, opened 2019-01-14
SG Optimizer Minify the HTML Output Breaks PDF


jakejackson1:
When enabled and the PDF is displayed in the browser, the SG Optimizer code will run anything captured in its buffer through a HTML minification parser which causes the PDF to become corrupted.

Issue #386, opened 2016-05-23
WP External Links Plugin Conflict


jakejackson1:
In 3.7.7 user had trouble generating PDF with the WP External Links plugin active (https://wordpress.org/plugins/wp-external-links/). Test it out and see if there is a problem.

TheWebmasterCo:
I can confirm this behavior in 4.0.4 -

The plugin author has a FAQ where he outlines some filters and methods to alleviate.

[https://wordpress.org/plugins/wp-external-links/faq/]

Using _wpel_apply_settings_ you might be able to work something in at runtime, but to ignore all links he is looking for a class or link attribute. Not sure how helpful those will be. As more time permits down the road, I might be able to revisit and try a few solutions.

jakejackson1:
A patch for this will be available in Gravity PDF 4.1.

PR #1556, opened 2024-07-23
Performance: Add temporary PDF cache

### LINKED ISSUE #1370: Partial Entries Add-on: Notification with PDF attached is cached
**Description**
When enabling the Partial Entries add-on on a form and setting up a Notification email with a Gravity PDF-attachment the generated PDF isn't removed from disk after the email is sent and this causes caching problems when the form actually does get submitted, as the old version is used. 

**Step To Reproduce**
1. Enable Partial Entries on a form
2. Setup a Notification email that is sent on the event "Partial Entries: Saved"
3. Setup a Notification email to be sent on the event "Form Submitted"
4. Setup a PDF to be attached to both Notification
5. In the form create a few fields that are required
6. Preview the form, fill out one of the fields and hit submit. 
7. When a validation error occurs, fill the rest of the form out and submit
8. Preview the PDF attachment from both emails sent

**Expected behavior**
The Form Submission PDF attachment should not use the cached version and instead correctly show all entered information on form submission.

### LINKED ISSUE #1537: Replace PDF cleanup after form submission
**Description**

Background processing emails is becoming more common, and auto-deleting the PDF attachments right after form submission can cause errors when the async email is sent. To correct this, replace the auto-cleanup logic with a scheduled event that runs 1 hour after the initial form submission. 

WP SMTP Mail v4 includes async mail sending and it should be possible to replicate the error with it.

### LINKED ISSUE #1539: Caching - add clear cache tool + dev mode bypass

jakejackson1:
## Description

Automatically save PDFs to disk for up to 1 hour, and return the cached copy on subsequent requests (if it exists). This is a smart cache that auto-detects if the form, entry, or PDF settings have been modified and will generate a new PDF if that's the case. 

Manually flushing the cache will be handled in a separate PR, alongside https://github.com/GravityPDF/gravity-pdf/issues/1527

This PR standardizes the PDF generation system so all PDF generation processes (view/download/notification/preview/bulk generate/API) flow through the same set of functions/methods to create the documents and save them to the temporary cache:

```
1. Model_PDF::generate_and_save_pdf()
2. Model_PDF::process_and_save_pdf()
3. Helper_PDF::generate()
4. Helper_PDF::save_pdf()
```

The calling function/method to `Model_PDF::generate_and_save_pdf()` will then determine what to do with the saved PDF. When viewing/downloading documents, the PDF will be read from disk and streamed to the browser with appropriate browser caching headers. When sending notifications, the absolute path to the PDF is passed to wp_mail() as an attachment, when using the API the absolute path is returned etc.

The `View_PDF::generate_pdf()` workflow to view/download PDFs is deprecated and should no longer be used directly, but for backwards compatibility will still work correctly.

## Testing instructions

This is a critical application path and should be thoroughly tested under all conditions to verify everything works as expected:

- [x] Viewing and Downloading PDF
- [x] Receive cached PDF when viewing and downloading PDF again
- [x] Updating form and viewing/downloading PDF for fresh copy
- [x] Updating entry and viewing/downloading PDF for fresh copy
- [x] Updating form PDF settings and viewing/downloading PDF for fresh copy
- [x] ?data=1 helper param
- [x] ?html=1 helper param
- [x] ?print=1 helper param
- [x] All view/download caching and URL options (above) with deprecated legacy PDF endpoint (from v3.0)
- [x] Set up multiple PDFs on a form and attach to multiple notifications. Verify cached copy is used when attaching the same PDF to 2+ notifications
- [x] Resend Notification feature uses cached PDFs when running multiple times
- [x] Resend Notification feature generates new PDFs if the form/entry/PDF settings are modified
- [x] New PDF generated after editing an entry via GravityView / Gravity Wiz / Gravity Flow
- [x] The `gfpdf_post_save_pdf` hook works as expected on form submission
- [x] The `gfpdf_override_pdf_bypass` hook bypasses the cache
- [x] The legacy `save` option works as expected on form submission
- [x] GPDFAPI::create_pdf() API leaves the cache intact when running
- [x] GPDFAPI::create_pdf() API uses cached copy on subsequent requests
- [x] Developer Toolkit PDFs view/download PDF 
- [x] Developer Toolkit PDFs support view/download caching and URL options (above)
- [x] Set up multiple Developer Toolkit PDFs on a form and attach to multiple notifications. Verify cached copy is used when attaching the same PDF to 2+ notifications
- [x] Developer Toolkit Legacy PDFs (with `advanced_template` option on): view/download + notifications (as above)
- [x] PDF Background Processing queue with standard and Developer Toolkit PDFs. Verify cached copy is used when attaching the same PDF to 2+ notifications
- [x] Manually calling `View_PDF::generate_pdf()` will view/download PDF as expected (including all URL params)
- [x] Legacy `gfpdfe_pdf_output_type` filter works
- [x] WP External Links plugin conflict: https://github.com/GravityPDF/gravity-pdf/issues/386
- [x] Weglot plugin conflict: https://github.com/GravityPDF/gravity-pdf/pull/1505
- [x] Siteground Optimizer minification conflict: https://github.com/GravityPDF/gravity-pdf/issues/863
- [x] Auto-purge PDF if associated template/config is updated
- [x] https://github.com/GravityPDF/gravity-pdf/issues/1403
- [x] https://github.com/GravityPDF/gravity-pdf/issues/1370
- [x] Gravity Perks Conditional Logic Date support on product fields
- [x] Gravity Perks Populate Anything support (live merge tags)
- [x] PDF Cached auto-cleaned up every hour by WordPress cron task
- [x] Async email sending e.g. WP Offload SES
- [x] Core Booster - Entry Notes
- [x] Gravity Perks Nested Forms - changes to nested form entries
- [x] PDF for GravityView works as expected
- [x] Enhanced Downloader
- [x] Bulk Generator works as expected
- [x] Previewer works as expected
- [x] PDF to Images works as expected

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/api.php
+++ b/api.php
@@ -448,19 +448,21 @@ final class GPDFAPI {
 	}
 
 	/**
-	 * When provided the Gravity Form entry ID and PDF ID, this method will correctly generate the PDF, save it to disk,
-	 * trigger appropriate actions and return the absolute path to the PDF.
+	 * Generate a PDF, save it to disk, and return the absolute path to the document
 	 *
 	 * See https://docs.gravitypdf.com/v6/developers/api/create_pdf/ for more information about this method
 	 *
-	 * @param  integer $entry_id The Gravity Form entry ID
-	 * @param  string  $pdf_id   The Gravity PDF ID number (the pid number in the URL when viewing a setting in the admin area)
+	 * @param int    $entry_id     The Gravity Form entry ID
+	 * @param string $pdf_id       The Gravity PDF ID number (the pid number in the URL when viewing a setting in the admin area)
+	 * @param bool   $bypass_cache Force a new PDF to be generated
 	 *
-	 * @return mixed            Return the full path to the PDF, or a WP_Error on failure
+	 * @return string|WP_Error   Return the full path to the PDF, or a WP_Error on failure
 	 *
 	 * @since 4.0
+	 * @since 6.12 All PDFs are cached on disk for ~1 hour, but are auto-purged if the form, entry, or PDF settings change
+	 *        Re-running the method will return the cached PDF if it exists, unless $bypass_cache = true
 	 */
-	public static function create_pdf( $entry_id, $pdf_id ) {
+	public static function create_pdf( $entry_id, $pdf_id, $bypass_cache = false ) {
 
 		$form_class = static::get_form_class();
 
@@ -478,16 +480,19 @@ final class GPDFAPI {
 			return new WP_Error( 'invalid_pdf_setting', esc_html__( 'Could not located the PDF Settings. Ensure you pass in a valid PDF ID.', 'gravity-forms-pdf-extended' ) );
 		}
 
-		$pdf  = static::get_mvc_class( 'Model_PDF' );
-		$form = $form_class->get_form( $entry['form_id'] );
+		if ( $bypass_cache ) {
+			add_filter( 'gfpdf_override_pdf_bypass', '__return_true', 9999 );
+		}
 
-		add_filter( 'gfpdf_override_pdf_bypass', '__return_true' );
-		do_action( 'gfpdf_pre_generate_and_save_pdf', $form, $entry, $setting );
-		$filename = $pdf->generate_and_save_pdf( $entry, $setting );
-		do_action( 'gfpdf_post_generate_and_save_pdf', $form, $entry, $setting );
-		remove_filter( 'gfpdf_override_pdf_bypass', '__return_true' );
+		/** @var \GFPDF\Model\Model_PDF $pdf */
+		$pdf         = static::get_mvc_class( 'Model_PDF' );
+		$path_to_pdf = $pdf->generate_and_save_pdf( $entry, $setting );
 
-		return $filename;
+		if ( $bypass_cache ) {
+			remove_filter( 'gfpdf_override_pdf_bypass', '__return_true', 9999 );
+		}
+
+		return $path_to_pdf;
 	}
 
 	/**
--- a/pdf.php
+++ b/pdf.php
@@ -1,7 +1,7 @@
 <?php
 /*
 Plugin Name: Gravity PDF
-Version: 6.11.1
+Version: 6.12.0
 Description: Automatically generate highly customizable PDF documents using Gravity Forms.
 Author: Blue Liquid Designs
 Author URI: https://blueliquiddesigns.com.au
@@ -28,7 +28,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 /*
  * Set base constants we'll use throughout the plugin
  */
-define( 'PDF_EXTENDED_VERSION', '6.11.1' ); /* the current plugin version */
+define( 'PDF_EXTENDED_VERSION', '6.12.0' ); /* the current plugin version */
 define( 'PDF_PLUGIN_DIR', plugin_dir_path( __FILE__ ) ); /* plugin directory path */
 define( 'PDF_PLUGIN_URL', plugin_dir_url( __FILE__ ) ); /* plugin directory url */
 define( 'PDF_PLUGIN_BASENAME', plugin_basename( __FILE__ ) ); /* the plugin basename */
--- a/src/Controller/Controller_Install.php
+++ b/src/Controller/Controller_Install.php
@@ -202,6 +202,8 @@ class Controller_Install extends Helper_Abstract_Controller implements Helper_In
 	 * Determine if we should be saving the PDF settings
 	 *
 	 * @since 4.0
+	 *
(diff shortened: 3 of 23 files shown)


jakejackson1:
TODO - add unit test for https://app.codecov.io/gh/GravityPDF/gravity-pdf/pull/1556/blob/src/Model/Model_PDF.php#L819

jakejackson1:
**TODO**

- [x] ~When entry metadata is modified the PDF cache is not cleared~ too volatile to check all. Need specific keys. Handle on case-by-case basis. 

- [x] Directly modifying the template/config PHP file doesn't clear the cache
- [x] PDF cache is not cleared if different user tries to view document. Breaks PDFs that reference {user} merge tag (or equivalent). 
- [x] Form object can be updated on form submission. There may be others, but $field->form_id and $field->enablePrice were dynamically added to field objects when processing a PDF. 
- [x] Editing the Gravity Perks Nested Forms entries doesn't clear the cache
- [x] When the Core Booster notes setting is enabled, adding new notes to an entry doesn't clear the PDF cache. Added `gfpdf_cache_hash_array` filter for CB to modify.
- [x] Unit test Cache class
- [x] [PDF to Images extension](https://github.com/GravityPDF/gravity-pdf-to-image/issues/9) does not currently work with new cache and will need to be updated

jakejackson1:
Core Booster support added https://github.com/GravityPDF/gravity-pdf-core-booster/pull/95

jakejackson1:
PDF for GravityView is not affected by the change, but doesn't currently make use of the temporary cache. Have opened issue about adding support https://github.com/GravityPDF/pdf-for-gravityview/issues/50

jakejackson1:
Enhanced Download support added https://github.com/GravityPDF/gravity-pdf-enhanced-download/pull/11

jakejackson1:
Bulk Generator is not affected by the change, but doesn't currently make use of the temporary cache. Have opened issue about adding support https://github.com/GravityPDF/gravity-pdf-bulk-generator/issues/184

jakejackson1:
Previewer not affected by the change, but doesn't make use of temporary cache. Have opened issue about it https://github.com/GravityPDF/gravity-pdf-previewer/issues/357

jakejackson1:
PDF to image compatibility added https://github.com/GravityPDF/gravity-pdf-to-image/pull/11 (still maintains independent image cache; a separate issue was created to combine the two).

PR #1563, opened 2024-08-29
Allow Form PDF Settings to be previewed before saving

### LINKED ISSUE #1528: PDF Preview Feature
**Description**
When setting up a PDF on a form, add a "Preview" button so the results can be previewed before the changes are saved. Along with improving the initial setup, it makes PDF settings updates more convenient since you can review the changes before they go live.

jakejackson1:
## Description

Add a "Preview PDF" button to the form PDF settings so users can preview the document with the current settings before saving.

## Testing instructions
1. Go to the Form Editor
2. Navigate to Form Settings -> PDF
3. Add a new PDF
4. Use the Preview PDF button
5. Make changes to the PDF settings
6. Use the Preview PDF button again
7. Save the PDF settings
8. Make changes to the PDF settings
9. Use the Preview PDF button
10. Repeat step 8-9 as many times as necessary 

## Screenshots 

CleanShot 2024-08-29 at 14 47 08@2x
CleanShot 2024-08-29 at 14 47 20@2x

## Checklist:
- [x] I've tested the code.

## Additional Comments 

The following extensions require a compatibility update:

1. Core Booster (filter fields)
2. Developer Toolkit (toolkit setting)


jakejackson1:
TODO:

- [x] Allow `entry` to be passed to the `get_items()` endpoint and only return PDFs that would be active for that entry (the PDF is set to active and the conditional logic passes).
- [x] Increase code coverage for API

PR #1653, opened 2026-05-15
ci: speed up wp-env across PHPUnit and Playwright workflows

jakejackson1:
Cache wp-env work directory, Composer, Yarn, Playwright browsers, and Docker layers across both test workflows so cold starts only pay the full provisioning cost on cache misses. Pin WordPress core to a fixed release in all wp-env configs (with a weekly bot that opens a PR when a new release ships) so the cache stays stable across runs.

PHPUnit: trim PR matrix to PHP 7.4/8.3/8.5 for fast feedback; full 7-version matrix runs on push to development/main and the new weekly cron. Drop the dedicated multisite matrix cell and run the multisite suite as a second step on the 8.3 cell against the same wp-env container.

Playwright: collapse the duplicate e2e-permalinks wp-env config into a single instance and flip permalink_structure per project via global-setup.ts; shard tests 4-ways in CI; aggregate shard results in a new notify job for the weekly schedule.


jakejackson1:
Too broad an update. Closing

PR #1664, opened 2026-06-05
Scope all PSR libraries and remove PSR Log compatibility layers

jakejackson1:
## Summary

Prefixes every PSR library under `GFPDF_Vendor` via php-scoper so the plugin is fully self-contained and can no longer collide with other plugins that ship `psr/log` v2 or v3. This lets us delete the runtime PSR Log 1/2/3 compatibility layer entirely.

## Changes

- **php-scoper config**
  - Dropped the `'Psr\*'` whitelist from `tools/php-scoper/config/monolog.php` so Monolog's PSR references get prefixed to `GFPDF_Vendor\Psr\*`.
  - Removed the `tools/php-scoper/config/mpdf.php` patcher that stripped the prefix back off `LoggerInterface` in the mPDF logger-aware traits.
- **Removed the PSR Log 1/2/3 compatibility layer**
  - Deleted `src/Helper/Log/MonoLoggerPsrLog2And3.php` (the v2/v3 proxy).
  - Stripped `Logger::get_monolog()`'s runtime version probe and `class_alias` bridging; `setup_logger()` now instantiates Monolog directly.
  - Deleted the obsolete `Test_MonoLoggerPsrLog2And3.php` and updated `Test_Logger.php`.
- **Type hints** — all 28 `use Psr\Log\LoggerInterface;` statements in `src/` now use `use GFPDF_Vendor\Psr\Log\LoggerInterface;`.

## Verification

- Regenerated `vendor_prefixed/` via `composer install` + `composer prefix`; confirmed Monolog and the mPDF traits consistently reference `GFPDF_Vendor\Psr\Log\*` and the autoloader maps the prefixed classes.
- Full PHP suite green: **1513 tests, 4862 assertions, 17 pre-existing skips**. The PDF-generation path (mPDF `setLogger` ← the prefixed logger) passes.
- PHPCS clean on the changed source.

## Dependent add-on PRs

Removing the unprefixed `\Psr\Log` shim means add-ons that type-hinted the logger against the plain interface now receive a scoped logger that no longer satisfies that hint. These companion PRs decouple each add-on's logger boundary so they work against both this (scoped) core and older (unprefixed) releases. They should land alongside this PR:

- GravityPDF/gravity-pdf-bulk-generator#201 — untype `$log` in `SessionId` / `ZipPath`
- GravityPDF/pdf-for-gravityview#88 — untype `$log` in `PdfEmail`
- GravityPDF/gravity-pdf-images#2 — untype `$log` in `DoesTemplateHaveGroup` + test logger
- GravityPDF/gravity-pdf-core-booster#125 — remove dead `Monolog\Logger` imports

An org-wide audit of the remaining ~30 dependent repos found no other add-on type-hinting the logger boundary (most consume it only via `Helper_Trait_Logger` + duck-typed calls, which is unaffected).

## Notes

`psr/container` is only a dev-time transitive dependency of php-scoper's own Symfony components and isn't referenced by any production package, so it's intentionally left out of scoping (it won't ship in the `--no-dev` release build).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jakejackson1:
Closing in favour of https://github.com/GravityPDF/gravity-pdf/pull/1696

PR #1674, opened 2026-07-14
Fix pre-rename text domain on three REST strings

thisismyurl:
Three REST argument strings in `src/Rest/Rest_Download_Pdf.php` still use the pre-rename `'gravity-forms-pdf-extended'` domain, while the plugin declares `Text Domain: gravity-pdf`. They never resolve to a translation as a result.

This points them at `'gravity-pdf'`, matching the correct sibling calls in the same file (lines 144/151/378). I left the legacy-slug identity checks in the updater alone, since those reference the old slug on purpose.

The REST download flow in Gravity PDF is thoughtfully built. Happy to help a few strings in it translate.

(full disclosure: AI helped me find the mismatch and confirm the updater checks were intentional; the change and this note are mine.)

### DIFF
--- a/src/Rest/Rest_Download_Pdf.php
+++ b/src/Rest/Rest_Download_Pdf.php
@@ -96,7 +96,7 @@ class Rest_Download_Pdf {
 			[
 				'args' => [
 					'entry' => [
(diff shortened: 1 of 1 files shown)


jakejackson1:
Thanks for the contribution @thisismyurl. Appreciate it.

PR #1695, opened 2026-08-14
Detect and warn about deprecated v3 functionality ahead of 7.0

jakejackson1:
## Summary

Gravity PDF 7.0 removes the v3 backwards compatibility layer, and today there is no way to tell which sites still depend on it. Legacy templates already render blank on Gravity Forms 3.0 and nothing says so: no fatal error, no warning, nothing in the debug log. The `gfpdfe_*` filters fire silently, so we don't know whether anyone still listens to them. The deprecated methods call `_doing_it_wrong()`, which never reaches WordPress' deprecation logging, and around twenty of them are empty stubs that emit nothing at all.

This PR detects the four kinds of leftover v3 usage a site can have and reports them in the Gravity Forms System Report, both WordPress Site Health surfaces, and a dismissible admin notice. It also swaps the deprecated methods and filters onto the real WordPress deprecation functions so they reach the deprecation log and Query Monitor, and restores the `RGForms` class that Gravity Forms 3.0 removed, which stops legacy templates rendering blank in the meantime.

Detection is a registry rather than a fixed list. `Deprecation` knows how to detect, group, record and warn about whatever its providers declare, and names neither the v3 layer nor the release removing it, so the next round of removals registers a provider class instead of being wired through the engine, model, view and controller. Everything detected here is grouped as deprecated: it all still works, and 7.0 is what removes it.

The work is admin-only apart from one write. The legacy download endpoint records the form it served, because a `?gf_pdf=1` link sitting in an email sent years ago is exactly the case scanning the forms table cannot find.

## Try it

```bash
yarn wp-env start

# Install a plain v3 template and a Business Plus one (the difference is the engine call)
yarn wp-env run cli wp eval 'file_put_contents( GPDFAPI::get_data_class()->template_location . "my-legacy.php", "<?php // a v3 template" );'
yarn wp-env run cli wp eval 'file_put_contents( GPDFAPI::get_data_class()->template_location . "my-business-plus.php", "<?php \$mpdf->AddPage();" );'

# Roll the recorded version back so the admin notices take a fresh detection
yarn wp-env run cli wp option update gfpdf_current_version 6.16.0
```

Then look at `Forms -> System Status`, `Tools -> Site Health` (both the Status and Info tabs), and any Gravity PDF admin screen for the notices. Delete the two files again and the rows clear on the next page load.

## Test plan

- [ ] `Forms -> System Status` shows a **Deprecated** section listing *Legacy Templates* and *Business Plus Templates* by file name, each linking to its upgrade guide
- [ ] `Tools -> Site Health` raises a recommendation, and the Info tab carries a matching section for a support ticket
- [ ] One notice lists every detected feature, and dismissing it clears all of them; a feature detected afterwards raises it again
- [ ] Deleting the template files clears every surface without needing a version bump
- [ ] A PDF using a v3 template renders with content on Gravity Forms 3.0 (blank on `development`)
- [ ] A `?gf_pdf=1` download still works, and adds its form to the report
- [ ] `yarn test:php`, `yarn test:php:multisite` and `yarn test:e2e` pass

<details>
<summary>More info</summary>

### Detection

`GFPDF\Statics\Deprecation` is a registry rather than a fixed list. It knows how to detect, group and warn about whatever its providers declare, and nothing in it names the v3 layer or the release removing it. `GFPDF\Statics\Deprecation_V3` is the one provider today, declaring four features and the version each goes in — so a later round of removals (a feature going in 7.1, say) registers itself instead of being wired through the engine, the model and the controller. The contract is `Helper_Interface_Deprecated_Features`.

Every feature is either **unsupported** (already stopped working) or **deprecated** (still works, scheduled to go). Every v3 feature is deprecated: 7.0 is what removes the layer, so nothing detected here has stopped working yet. `Deprecation::get_groups()` drops a group nothing declares, so no surface carries an empty section around and the vocabulary is in place for a later round without a section reading *None detected* until one arrives. The engine owns the group list and its order; the view names them, once, for all three surfaces. `Deprecation::refresh_signals()` pairs detection with recording, so both Site Health surfaces call the engine directly rather than reaching through the system report's model for a detection that has nothing to do with the report.

| Feature | Group | How |
|---|---|---|
| **Legacy Templates** | Deprecated | Any template `Helper_Templates` classifies in the `Legacy` group, which is one carrying no file headers (core templates declare `Group: Core`, so they're never flagged). |
| **Business Plus Templates** | Deprecated | The same set as above, split on whether the file calls `$mpdf->`. Driving the PDF engine is the one thing the Tier 2 add-on gave a template and a plain v3 template never had, so the call is what tells them apart. Reported under their own label, linking to the part of the guide that covers them — a plain v3 template is rebuilt from the existing developer docs, while one of these is a Bespoke build being ported to the Developer Toolkit. |

Both template rows read the template files and nothing else. An earlier revision also walked `gf_form_meta` for PDFs set to `advanced_template => Yes` and reported whatever template they pointed at, which meant toggling that setting on a Core template listed *zadani.php* in the report. A template is now reported for what it is rather than for how one PDF happens to be configured, and the detection loses a DB query. `is_advanced_template_pdf()` stays — the render path still branches on the setting.
| **Legacy Download URLs** | Deprecated | Two sources, unioned in one query. A `LIKE` for `gf_pdf=1` across the three columns Gravity Forms splits a form over finds a form *handing the URL out* — these are hand-written into a confirmation, notification or field, so the scan finds one nobody has followed yet. The endpoint itself records the form it served, which finds a URL living somewhere the scan can't see: a page, or an email that has already gone out. Matching `gf_pdf=1` rather than `gf_pdf=` keeps the scan to the value that actually routes to the legacy endpoint. |
| **Actions and Filters** | Deprecated | Walks `$wp_filter` for third-party listeners on the `gfpdfe_*` hooks, the v3-shaped `gfpdf_*` aliases (`gfpdf_orientation`, `gfpdf_privilages`, …) and the three `gfpdf_legacy_*` hooks. Walking rather than checking a fixed list picks up dynamic hooks like `gfpdfe_pdf_template_{form_id}`. Core's own `PDFRender::prepare_ids` registration is excluded via a shared constant. |

Trashed forms are skipped throughout — they aren't in the user's form list, so there's nothing to act on.

### Surfacing

**System Report** gains a **Deprecated** section at the head of the Gravity PDF Environment block. Rows follow the report's own condensed convention: the detections on one line, then the cross and a red message, which Gravity Forms renders itself from the `is_valid` and `validation_message` keys rather than us building the markup. Legacy templates are listed by file name alone, since the report already states the PDF working directory, and each form that served a legacy URL links to its own PDF settings, which is where those URLs get replaced. Sections with no items are dropped, which is how these stay hidden on a clean site.

Sections and their rows are matched by name rather than by position. Two sections at the head would otherwise have renumbered the four below them, and every section added after would do it again — which also moves the ground under `gfpdf_system_status_report_items`, a public filter keyed by that same index. The two group sections are built from `View_System_Report::get_deprecated_groups()`, so a group is named once for all three surfaces instead of again in the model.

Legacy templates moved here out of **Outdated Templates**, where the first pass of this branch put them — the two conditions have entirely different remedies and reading as one run-on warning wasn't helping anyone.

**Site Health** gains a test reporting the same detections, grouped, with a link to the system report. This is the surface that can't be permanently dismissed: it clears itself once the site stops using the deprecated functionality, returns if it starts again, and carries onto the Dashboard through the Site Health Status widget. The Info tab gains a matching section per group, so the detections travel in the debug export users paste into a support ticket — always present, reporting "None detected" on a clean site rather than leaving the reader to guess whether the check ran. Both are gated on `gravityforms_view_settings`, so form titles, PDF names and template paths aren't exposed to administrators without Gravity Forms access.

Site Health has room the one-line report row doesn't, so it names a template's full path where the report abbreviates to the file name.

**One admin notice** listing every detected feature, raised through the existing one-time action routes. One rather than one per feature: four warnings stacked on the same screen is noise a reader learns to scroll past, and they all say the same thing — go and look at the system report.

Dismissing it records each feature it was listing, rather than the notice itself, so silencing today's list says nothing about the next round of removals: a feature detected later was never on it and raises the notice again on its own. That needed a `dismiss` key on the route, which `route()` now honours the way it already honours `process`; a route supplying one owns suppressing itself through `condition`, since `route_notices()` still reads the action's own dismissal. Each line uses `Deprecation::get_feature_notice()`, the same sentence the report uses, which derives from the registration's group, label and version rather than living in either view. A registration can also declare a `notice` of its own, which is what the hooks feature needs: its label is a category, so "Support for Actions and Filters will be removed" reads as the whole hook API once it is out of the report row that lists the hook names underneath it. It says "Code on this site uses Gravity PDF hooks that are removed in version 7.0." instead, on every surface. That also moved the report's two per-feature copy overrides into the registrations, so the notice and the report can no longer drift. Functionality that has already gone reads as an error, the rest as a warning, and the button leads to the system report where the detections are listed. It reaches Network Admin as well — `Helper_Notices` already switches to `network_admin_notices`, and the network dashboard and plugins screens both match the pages a notice may appear on. What it reports there is the main site, which is the scope `gfpdf_settings` and the detection record live at; a subsite raises its own notice in its own admin. Aggregating a whole network is deliberately out of scope for 6.17.

The notices don't detect anything themselves — they read what the last detection recorded. Detection walks the form table and the template directory, which is far too much to repeat on every admin page a notice can appear on. `Controller_Upgrade_Routines` takes that record at install and, ungated, on every version change — a release being both when a new round of removals arrives and the only moment the notices need to learn of one. The report and Site Health screens refresh it as they go, since they detect live regardless, so a site that fixes everything stops being told about it without waiting for the next release. The record and the dismissals beside it live in `gfpdf_settings`, which the uninstaller already drops whole.

**The legacy endpoint records the form it served**, in a `gfpdf_legacy_endpoint_usage` option of its own. It is written after the request resolves to a real PDF, never before — `$_GET['fid']` is anonymous input, and recording it unvalidated would let any visitor write arbitrary IDs into the option. It is written before the PDF is served, though, so a request the middleware turns away still counts as one made: the signal is that the URL is out there being used. Each form is stored against the time it was last served and expires 30 days later: Gravity PDF can't tell whether a URL it served last year still exists, and without an expiry a form whose links were fixed months ago would be reported forever. The stamp is refreshed at most once a day, so a link still in use never reaches the expiry while one nobody follows drops off on its own, and the front end writes about as often as it did when only the first request counted. Expired entries are pruned by the read, on the admin screens that check them. A write also adds the feature to the notices' record, which is otherwise only taken at install and on each version change — a URL followed since then would go unmentioned until the next release. The option is deliberately not in `gfpdf_settings` and not autoloaded: it is written from the front end and read only by the report, Site Health and the uninstaller. `Model_Uninstall` doesn't name the option: `Helper_Interface_Deprecated_Features::get_stored_options()` declares what a provider stores and `Deprecation::delete_stored_data()` fans the delete out across the registry, so a later round of removals brings its own storage with it rather than needing a line added to the uninstaller.

**A shim for the class Gravity Forms 3.0 removed.** `RGForms` was an empty subclass of `GFForms`, deprecated in GF 2.10.5 (`@remove-in 3.0`) and removed in 3.0. Every v3 template Gravity PDF ever shipped opens with `if ( ! class_exists( 'RGForms' ) ) { return; }`, so on GF 3.0 those PDFs render with no content at all — no fatal, nothing in the log, a ~1.1KB blank page and no way for the site owner to find out why. Verified against GF 3.0.2. `Deprecation_V3::restore_v3_form_class()` aliases `GFForms` back to the old name when a legacy template is about to render, called from `Helper_PDF::set_template()` and from `Model_PDF::handle_legacy_tier_2_processing()`, which runs the template file itself and never reaches `set_template()`. Detection is unchanged, but the group isn't: a legacy template renders correctly again, so it is reported as **Deprecated** — it works today and goes in 7.0, which is what that group means. Worth a reviewer's eye — it re-declares a name Gravity Forms deliberately dropped, guarded on `class_exists()` both ways, and scoped to the legacy render path rather than declared globally.

**Log warnings** on Advanced Templating processing and any `gfpdfe_*` filter that has a listener, so support can identify an affected site from a submitted log file without a back-and-forth. The filter warning is written once per hook per request — otherwise `gfpdfe_pdf_template` would write a line for every PDF field. Rendering a legacy template writes nothing: all three report surfaces already name the file, so a line per PDF told a reader nothing they couldn't see there, and `View_PDF::autoprocess_core_template_options()` stays the styling decision it was.

### Real deprecation notices

- `_doing_it_wrong()` → `_deprecated_function()` / `_deprecated_argument()` across the legacy-endpoint, v3-template and stale-stub surface. These now fire `deprecated_function_run` and land in WordPress' deprecation log and Query Monitor. The bespoke prose messages collapse into the `$replacement` argument.
- The ~20 silent stubs (`Model_Settings` font methods, `Model_Install` uninstall proxies, `Helper_Notices`, the `pdf.php` canonical-release notices) now emit a notice instead of returning quietly. None of them are hooked, so nothing fires on a normal request.
- The v3 shim classes in `src/deprecated.php` warn from their implemented methods, not just the `__call` fallback — `PDF_Common::get_ids()` and friends previously warned about nothing, and the `mPDF` shim constructor means a v3 template calling `new mPDF()` is no longer silent. `GFPDF_Core::setup_constants()` and `PDFRender::prepare_ids()` are deliberately excluded: core calls those itself.
- The `gfpdfe_*` filters route through `Deprecation::apply_filters()`, a thin wrapper over `apply_filters_deprecated()` that takes the replacement and the removal version from whichever registered feature owns the hook. It early-returns when nothing is listening, so the PDF render path (~16 call sites) pays nothing on sites that have already moved on.

### Helper_Notices fixes

Two fixes came out of building the notice, both on the pre-existing `view_class` route key that any add-on can already use:

- A caller-supplied `notice-*` class now replaces the default `updated` state rather than joining it — `div.updated` out-specifies `.notice-warning`, so an appended warning still rendered green.
- Queued notices are stored as message/class pairs rather than keyed by their class, which silently dropped a second notice sharing the same one.

`Controller_Actions::route_notices()` also returns early on the admin pages where `Helper_Notices` discards the notice anyway, so route conditions no longer run on every admin request.

### Notes

- **`gfpdfe_signature_width` already has a v4 replacement** — `gfpdf_signature_width` is fired immediately after it in `Field_Signature.php`. The plan lists creating one as an open question; it isn't needed.
- A dismissal is permanent *for that feature*. Fix your legacy templates, dismiss the notice, install another one next year and it won't return — Site Health will, and can't be dismissed at all, which is what makes the per-feature dismissal defensible.

### Testing

- `Test_Deprecation` covers the engine, including a fake provider declaring a feature removed in 7.1 — which is the whole of what registering a later round of removals amounts to. `Test_Deprecation_V3` covers the v3 detectors, with forms from a new `CreatesLegacyDownloadUrls` concern shared with the system report tests — including the endpoint recorder: the union of both sources, write-once, the option staying out of `wp_load_alloptions()`, a trashed recorded form dropping out, and `gf_pdf=0` no longer reading as a legacy URL. `Test_Controller_PDF` covers a request to the endpoint recording the form it served.
- `Test_Controller_System_Report` covers each signal, the *Enable Advanced Templating* setting reporting nothing on its own, a Business Plus template taking a row of its own, every surface naming a template by file rather than by path, and the registration, capability gate and reported detections of both Site Health surfaces. It looks report sections up by the ID they declare rather than by position, so adding or dropping one doesn't move every assertion below it.
- `Test_Controller_Actions` covers the one route the features share, its severity, and the full dismissal cycle through rendered HTML: both features named in the notice, gone after dismissing, then a feature detected afterwards bringing the notice back naming only itself. It also pins that the notice reaches Network Admin. `Test_Controller_Upgrade_Routines` covers the record being taken at install and on a version change, and cleared on the next release once the site is fixed. `Test_Uninstaller` pins that the record and the dismissals go with `gfpdf_settings`, and that the endpoint's own option is deleted beside it. `Test_Deprecation` pins the fan-out itself, through a fake provider with storage of its own.
- `setExpectedIncorrectUsage()` → `setExpectedDeprecated()` wherever the primitive changed; tests that reach a deprecated path incidentally now declare it.
- New Playwright coverage with Chromatic snapshots of four new screens: the system report sections, the Site Health test expanded, the Site Health Info tab panels, and one of the admin notices. Chromatic archives the whole document and has no element-level crop, so `isolateForSnapshot()` hides everything but the target and its ancestors first. The detected form ID is swapped for a constant before each snapshot, since it differs between environments.
- Every Chromatic snapshot in the suite now goes through one `snapshot()` helper that holds until the page stops moving — fonts loaded, local images complete, geometry unchanged across two polls — because the archive is taken at the call and whatever is mid-flight is what gets baselined. This is why the existing snapshot specs are touched. It fixes three sources of baseline churn that predate this branch: WordPress relocating admin notices after `.wp-header-end` on jQuery ready, and the entry list rendering `date_created` to the minute (the two specs that snapshot that screen pass a fixed date, which GFAPI takes verbatim; it is opt-in rather than the factory default, since an entry's creation time is also what times a logged out visitor's PDF access). The admin bar's Gravatar is aborted in the fixture: wp-env cannot reach that origin, so whether the avatar resolved was down to network luck.
- `switchToCodeEditor()` waits for TinyMCE to finish initialising before clicking the tab. WordPress' `switchEditor()` calls `editor.hide()` and copies the TinyMCE iframe's height onto the textarea it reveals; instrumenting the click found the editor uninitialised on four runs in five, where `hide()` throws above the line that swaps `tmce-active` for `html-active` — leaving the editor in visual mode with its textarea hidden, which is the intermittent `Notifications` failure. The same moment fixes the textarea's height, so this also settles the rich text boxes that moved between snapshots; a height baked in at the switch cannot be settled afterwards.
- Integration suite: 1616 tests, 0 failures, 46 skipped. Multisite: 1616 tests, 0 failures. The deprecation spec now runs in a Playwright project of its own, `core-isolated`, chained between `core` and the permalink pass: its fixtures install site-wide signals, and the notice they raise renders on every admin page the suite's other full-page snapshots are taken on, so `fullyParallel` running it beside them put the notice into an unrelated Chromatic baseline. That spec has not been re-run since it moved projects, so CI is the first pass on it there. The E2E spec installs one template of each kind for the run — a headerless file, and one calling `$mpdf->` — so both template rows appear in the snapshot rather than being covered only by PHPUnit. `composer run phpcs:lint .`, `composer run phpcs:compatibility .`, `yarn lint:js` and `yarn lint:css` all clean.

### Follow-ups (not in this PR)

- Publish the upgrade guides before release: `/upgrade/legacy-templates/`, `/upgrade/legacy-download-urls/` and `/upgrade/deprecated-filters/`. All are linked from the report, Site Health and the notices, and none exist yet. The Advanced Templating row links into the legacy templates guide, since a PDF using the add-on is always a legacy template as well. Drafted in [gravity-pdf-documentation#204](https://github.com/GravityPDF/gravity-pdf-documentation/pull/204).
- Tier 2 add-on customers need a commercial conversation, not just a notice — flag to support before 7.0 ships.
- Changelog entry — there's no 6.17.0 section yet, and this repo writes those at release time.

</details>

### DIFF
--- a/src/View/View_System_Report.php
+++ b/src/View/View_System_Report.php
@@ -5,6 +5,7 @@ declare( strict_types=1 );
 namespace GFPDF\View;
 
 use GFPDF\Helper\Helper_Abstract_View;
+use GFPDF\Statics\Deprecation;
 
 /**
  * @package     Gravity PDF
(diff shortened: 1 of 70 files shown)


src/View/View_System_Report.php
```diff
@@ -146,6 +147,230 @@ public function get_template_check_message( string $path, string $template_versi
...
+		if ( $signals === [] ) {
+			return $result;
+		}
+
+		$labels = Deprecation::get_feature_labels();
+
+		$result['status'] = 'recommended';
+		$result['label']  = sprintf(
+			/* translators: %s: The Gravity PDF version the functionality is removed in */
+			esc_html__( 'Your site uses Gravity PDF functionality that is removed in version %s', 'gravity-pdf' ),
```
jakejackson1: "... that will be removed in version %s"

src/View/View_System_Report.php
```diff
@@ -146,6 +147,230 @@ public function get_template_check_message( string $path, string $template_versi
...
+			esc_html( Deprecation::REMOVED_IN )
+		);
+
+		$result['description'] = '<p>' . sprintf(
+			/* translators: %s: The Gravity PDF version the functionality is removed in */
+			esc_html__( 'The following functionality dates back to Gravity PDF version 3 and is removed in version %s. Update it before then, or the PDFs that rely on it will stop working.', 'gravity-pdf' ),
```
jakejackson1: "before then, so the PDFs that rely on it will continue working."

src/View/html/Actions/deprecated_feature.php
```diff
@@ -0,0 +1,48 @@
...
+	<strong>
+		<?php if ( $args['unsupported'] ) : ?>
+			<?php
+			/* translators: %s: The name of the deprecated feature, such as Legacy Templates */
+			printf( esc_html__( 'This site uses Gravity PDF functionality that is no longer supported: %s.', 'gravity-pdf' ), esc_html( $args['feature']['label'] ) );
+			?>
+		<?php else : ?>
+			<?php
+			/* translators: 1: The name of the deprecated feature, 2: The Gravity PDF version that removes it */
+			printf( esc_html__( 'This site uses Gravity PDF functionality that version %2$s removes: %1$s.', 'gravity-pdf' ), esc_html( $args['feature']['label'] ), esc_html( $args['feature']['removed_in'] ) );
```
jakejackson1: This site uses Gravity PDF functionality that will be removed in version %2$s:

src/View/html/Actions/deprecated_feature.php
```diff
@@ -0,0 +1,48 @@
...
+			?>
+		<?php endif; ?>
+	</strong>
+
+	<br>
+
+	<?php if ( $args['unsupported'] ) : ?>
+		<?php esc_html_e( 'The PDFs relying on it have stopped working.', 'gravity-pdf' ); ?>
+	<?php else : ?>
+		<?php esc_html_e( 'It still works today, and needs replacing before that release to keep your PDFs working.', 'gravity-pdf' ); ?>
```
jakejackson1: It still works today, but needs to be replaced before the %s update to keep your PDFs working.

src/View/View_System_Report.php
```diff
@@ -146,6 +147,357 @@ public function get_template_check_message( string $path, string $template_versi
...
+	 * @param array $feature The feature's registration
+	 *
+	 * @since 6.17.0
+	 */
+	protected function get_deprecated_filter_feature( array $filters, array $feature ): array {
+		$lines = [];
+
+		foreach ( $filters as $name => $count ) {
+			/* translators: 1: Filter name, 2: Number of callbacks attached to it */
+			$lines[] = sprintf( __( '%1$s has %2$d listener(s)', 'gravity-pdf' ), $name, $count );
```
jakejackson1: Use `_n()`

src/View/View_System_Report.php
```diff
@@ -146,6 +147,357 @@ public function get_template_check_message( string $path, string $template_versi
...
+	 * with the name alone. Deprecation::get_groups() decides which of them a surface sees.
+	 *
+	 * @return array<string, array<string, string>>
+	 * @since 6.17.0
+	 */
+	public static function get_deprecated_groups(): array {
+		$names = [
+			Deprecation::GROUP_UNSUPPORTED => [
+				'label'       => __( 'Unsupported', 'gravity-pdf' ),
+				'description' => __( 'These features have been removed and the Gravity PDF documents that rely on them will stop working.', 'gravity-pdf' ),
```
jakejackson1: These features have been removed and any Gravity PDF document that relied on them will stop working.

src/View/View_System_Report.php
```diff
@@ -146,6 +147,357 @@ public function get_template_check_message( string $path, string $template_versi
...
+	 */
+	public function get_deprecated_features_test( array $signals ): array {
+		$result = [
+			'label'       => esc_html__( 'Your site does not use deprecated Gravity PDF functionality', 'gravity-pdf' ),
+			'status'      => 'good',
+			'badge'       => [
+				'label' => esc_html__( 'Gravity PDF', 'gravity-pdf' ),
+				'color' => 'blue',
+			],
+			'description' => '<p>' . esc_html__( 'Gravity PDF is removing functionality that is now deprecated. Nothing on this site relies on it.', 'gravity-pdf' ) . '</p>',
```
jakejackson1: Re-write to make the happy path clear and positive to the reader.

src/View/View_System_Report.php
```diff
@@ -146,6 +147,357 @@ public function get_template_check_message( string $path, string $template_versi
...
+		];
+
+		if ( $signals === [] ) {
+			return $result;
+		}
+
+		$result['status'] = 'recommended';
+		$result['label']  = esc_html__( 'Your site uses deprecated Gravity PDF functionality', 'gravity-pdf' );
+
+		$result['description'] = '<p>' . esc_html__( 'Your site is using Gravity PDF functionality that is being removed. Each item below links to instructions for replacing it.', 'gravity-pdf' ) . '</p>';
```
jakejackson1: "... links to instructions to upgrade."

PR #1722, opened 2026-09-07
The 7.0 font subsystem: remove the core installer, add language packs and the new Font Manager

jakejackson1:
> **Draft, tracking branch.** This is the long-running 7.0 font epic, not a finished change. It is here to make the work visible and reviewable in pieces; the plan splits it into separate PRs and this branch carries them together. Not for merge in its present shape.

## Summary

Gravity PDF 7.0 removes the core font installer — the one-click downloader that pulled ~80 font files from a GitHub mirror into the uploads directory, and the admin nag that asked for it. In its place: a small set of fonts bundled with the plugin so a fresh install renders properly offline, and language packs installed on demand from our own update origin.

**The installer is now gone.** Its AJAX handler, its manifest, the React bundle behind the Tools-tab button, and the E2E stubs that stood in for the download have all been deleted, along with the render-time directory glob that quietly registered whatever it left behind.

Deleting the manifest is the interesting part. `build/payload/core-fonts.json` was the only record of what the installer could put on disk, so removing it would leave an upgraded site with a fonts directory nothing recognises — every PDF that named one of those fonts would fall back to the bundled one. The same change therefore freezes what the manifest knew: 39 mPDF font families over 70 files, each face carrying its filename, its size and the hash the manifest already held. The 7.0 upgrade reads that, verifies each file byte-for-byte, and re-registers what is actually there under the names 6.x used. A site rendering Korean through `unbatang` keeps rendering it, with no network and no template edits — even though a fresh install of the Korean language pack will ship a different face.

Behind that, fonts moved out of the settings blob and into three tables read and written through one repository. `custom_fonts` used to be an autoloaded option every WordPress request unserialised whether or not a PDF was generated, and every font write read-modify-wrote it in full, so a background install and a settings save could clobber each other. Writes are now row-level, the load path performs no filesystem operation at all, and one cached read sits behind both mPDF registration and the Font Manager, so what renders and what the UI lists cannot drift.

`Model_Custom_Fonts` remains as a façade with its 6.x array shape intact, because `GPDFAPI::get_pdf_fonts()`, `Helper_Data::customFontData` and third-party add-ons read that shape.

**What replaces the installer now exists too.** A signed catalogue of font sources syncs from our own update origin; seventeen language packs install on demand from it, triggered by the site locale, by PDF settings, or by a render that meets a script it has no font for — that last one fetching inline, while the document draws, so a Japanese submission on a site with no Japanese pack produces a correct PDF rather than boxes. Installs run on their own background queue and report progress without touching the network. A health check notices a site whose fonts are wrong before its customers do, and the System Report carries the subsystem.

The catalogue's own integrity rules tightened along the way: an entry that omits the hash or size every download is checked against is now refused at sync, where a build mistake belongs, rather than surfacing mid-install on a customer's site. And a pack that renames a font key can now carry the old name with it, so a 6.x template naming the old one keeps rendering in the right face instead of silently falling back to the document font.

The Font Manager is rebuilt on `@wordpress/*` packages: a catalogue browser with server-side search, per-entry install with style selection, an updates panel, and — for sites that cannot reach the font server at all — **an offline import that takes a package zip**, verifying every file against the catalogue row before anything is written. Multisite is handled throughout: fonts are network-global, and one site cannot delete the whole network's files.

**The Font Manager now reads nothing from fixtures.** Its language settings were the last screen still backed by in-browser stand-ins; they save to the site now. An admin sets the language a PDF is assumed to be written in, the script it is typeset in, whether packs install themselves, and — row by row — which font should draw each language or script that is *not* the document's own. A single PDF can override the document language for itself.

Pinning the behaviour that table describes turned up two bugs older than this work. The document-script setting looked up a constant mPDF does not define, so every value it has ever been given was silently ignored and the script fell back to whatever the document language implied. And an unanswered dropdown on the PDF settings screen was stored as an empty array rather than an empty string — which is a `TypeError` in `strtoupper( $settings['pdf_size'] )` under PHP 8, and a literal `font-family: Array` in the generated stylesheet.

## Try it

On a site that used the core font installer, upgrade to this branch, **disconnect it from the internet**, and load any admin page.

```
wp option get gfpdf_db_version
wp db query "SELECT font_key, label, source FROM wp_gravitypdf_font"
```

Every font the installer had written should be listed. Generate a PDF that uses one of them — it should render exactly as before. The Tools tab should no longer offer a "Download Core Fonts" button, and the nag asking for it should be gone.

On a site that never used the installer, a fresh install should generate the default template for an English entry with true bold and italic, and an embedded ✔ on a Consent field, with no fonts on disk and no outbound HTTP.

**The Font Manager.** Open any PDF's settings and click the button beside the Font field. The catalogue will not have synced yet, so *Add new font → Browse Language packs* should offer a Refresh rather than an empty grid; refresh it, then install a pack and watch the progress land file by file.

**A render that installs what it needs.** With the Japanese pack *not* installed, submit an entry containing Japanese text and generate its PDF. The glyphs should be there, not boxes.

**The language settings.** In the Font Manager, *Language settings*. Change the font for a language, add one the table does not list yet, save, then reopen the manager — both should have survived, and a row you reset should lose its reset button. Then set *Default document language* to Japanese on a site with the Japanese pack installed and a font of your own chosen for the PDF: Japanese text should render in **your** font, because it is the document's own language. Set it back to English and the pack takes those runs back.

**The offline import.** *Add new font → Install from file* takes a pack zip. On a site that has never synced there is no catalogue to verify against, so it should refuse with a 404 that says so and point at the offline install guide — that refusal is what the new E2E test covers.

## Test plan

`yarn test:php` — **2,334 tests, 7,849 assertions**, passing. `yarn test:php:multisite` — **7,963 assertions**, passing. `yarn test:js` — **389 tests across 47 suites**; coverage 89.6 / 85.1 / 87.3 / 89.4 against the 75% threshold. `phpcs`, `lint:js` and `lint:css` clean.

Playwright: the `core` project is **65 passed, 3 failed** on a full run. Two of the three — `multiple-pdfs` and the license tab's invalid-key message — pass in isolation and failed only while three Jest coverage runs were competing for the machine. The third is `page-confirmation`, which fails identically with this whole branch stashed: a timeout inside `@wordpress/e2e-test-utils-playwright`'s `setPreferences`, unrelated to anything here. The Font Manager spec is now 13 tests and carries three Chromatic snapshots.

**This branch is rebased onto #1727, which unblocked `pdf-settings.spec.ts`.** Before that fix the spec failed 20 of its 21 tests here — every failure a 30 s timeout creating the form in `beforeEach`, because 821 forms shared its title and Gravity Forms de-duplicates titles in O(n²) against the whole table. It now runs 21/21 in 1.5 minutes, against 10.3 minutes of timeouts before.

One PHPUnit test is intermittently flaky under load: `Test_Render_Font_Trigger::test_a_japanese_submission_installs_the_pack_it_needs_before_it_draws`, which failed once in three full runs and passes 9/9 when the class is run alone. Worth pinning down before release rather than at review time.

`yarn test:js` passes clean, but under `--coverage` a handful of component suites time out non-deterministically — different ones each run, several of them suites this branch never touched. Treat a coverage-run failure as machine load rather than signal; the coverage figures themselves are collected regardless.

The adoption tests are not mocked. Four of the repo's existing PHPUnit font fixtures turn out to be byte-identical to their manifest entries, so they exercise the real hashes and independently corroborate the generated map. Neutering the verification fails two cases.

**Two behaviour changes to note before release:**

- Fonts dropped straight into the uploads fonts directory are no longer registered on every render. Existing ones are imported into the Font Manager during the upgrade, keeping the name templates use; new ones must go through the Font Manager or `GPDFAPI::add_pdf_font()`.
- The tables have columns, so a key outside the documented 6.x shape no longer survives a round trip through `update_font()`.

<details>
<summary>What's on the branch, and what's still to come</summary>

**Landed here**

- **Phase 0 spikes.** The mPDF fork (`GravityPDF/mpdf`, carrying the font-registry decoupling), the two registry layers against it, release-build size, script-detector cost, the Noto replacements for the licensing-flagged fonts, the language-pack split, file hosting, and a dry run of the Google Fonts pipeline.
- **Bundled fonts.** Four dehinted Arimo statics plus a DejaVu symbol subset, at the plugin root. `fonts/` adds 863 KB compressed; the zip is 5.16 MB against a 5.6 MB budget.
- **Phase 1a.** mPDF switched to the fork.
- **Phase 1b.** The font tables.
- **Phase 1c.** The registry, the package layers, the loose-font importer.
- **Phase 2.** Deleting the installer.
- **Phase 3.** The catalogue table and source registry, the repository, the signed sync and its pinned key, the streaming downloader, the install queue, the REST routes, the install triggers (locale, settings, and inline-at-render), the health check and System Report section, and multisite ownership.
- **Phases 4–5.** The Font Manager rebuilt on `@wordpress/*`: catalogue browser, entry install with style selection, updates panel, language settings, previews, and the offline package import (`POST /fonts/import`) with its UI.
- **Phase 6.** The language settings served for real: `GET`/`POST /fonts/settings`, the four §4.8 keys, the per-PDF `pdf_language`, and the §4.4 behaviour table pinned against a real mPDF.

**Phase 1b — the tables**

`Font_Schema` owns `gravitypdf_font`, `gravitypdf_font_file` and, on multisite, `gravitypdf_font_site`, created by `dbDelta` and verified with `SHOW TABLES` afterwards. The stored version carries an `-ms` suffix once the site table exists, so a single site converted to a network *after* upgrading mismatches and creates it — an `is_multisite()` check behind an equal-version compare never would. A failed pass logs `dbDelta`'s own messages and sets an hour-long transient, so a host that revoked `CREATE TABLE` retries hourly rather than per request.

`Font_Repository::all()` is two statements, cached under a `last_changed` stamp every write bumps, memoised per request, and merged in PHP with this site's visibility toggles on multisite. `verify()` is the only pass that touches the disk, and it is one `scandir()` plus one `filesize()` per distinct path, driven from the Font Manager and the health check rather than a render.

`Font_Migration` matches on the set of files a record points at rather than the key derived from it, so a retried pass mints nothing and — on multisite, where the tables are network-global — the first site to migrate a key owns it. **`custom_fonts` is deliberately left in place** as a frozen snapshot: deleting it was the one irreversible step in the upgrade, and the rollback path needs it. 7.0 never reads it after migration and never writes it.

`ensure_ready()` runs before the repository's first query on any request, gated by the autoloaded version option, because the upgrade routine only fires on an admin page load — a site that auto-updates and serves PDFs by REST or cron would otherwise render before it. Its mutating half runs under a network-level lock, and the version is written only after every step succeeds, so a failed pass is retried rather than closed by a version bump.

Unlinking is rows-first everywhere. A file is removed only once no surviving file row records its path, because two installs of one entry can share files.

**Phase 1c — the registry**

`Registry` builds both mPDF package layers from one cached read. Order is load-bearing: `FontRegistry::add()` prepends and mPDF reads the result in that order, so the layer added last is read first. The bundled layer therefore goes in second, and its `backupSubsFont` entries sort ahead of the rows'.

`Helper_PDF::begin_pdf()` builds the whole font config from the registry rather than from mPDF's built-in defaults, which the fork empties. `autoScriptToLang` is a 7.0 requirement rather than a preference — mPDF never substitutes Arabic or Indic glyphs, so a lang tag on the run is the only route to a font for those scripts. `useKerning` likewise: Arimo carries no legacy kern table, so GPOS is its only route to kerning.

`tools/mu-plugins/mpdf.php` was deleted here, and it mattered more than its size suggests: it merged `mode => 'c'` into every test-suite mPDF construction, forcing `onlyCoreFonts = true`, which meant **no test had ever proved a real font was embedded**. `Test_Bundled_Render` is the replacement guard, and it asserts `onlyCoreFonts === false` so a re-introduced stub fails loudly.

**Phase 2 — deleting the installer**

`Legacy_Installer_Files` is grouped by font family rather than by file because that is how the fonts were registered: `dejavusanscondensed` is four files under one key, and adopting them as four rows would change the key a template resolves. `use_otl` and `use_kashida` come across too — they are what shaped the text in 6.x.

The keys came from upstream mPDF's `FontVariables` at the revision before the font-package split. The fork's copy is empty, and its packages have replaced ten of these families outright, so no live source still has them.

Verification uses the Git blob hash the manifest already carried, not a fresh SHA-256. A SHA-256 could only be derived by re-downloading 80 MB today, which would attest to that download rather than to what the installer shipped — weaker provenance, not stronger. It is checked by streaming, size first, so a 12 MB font is never read whole to reject it on length.

`Font_Repository` gained a `Font_Population_Pass` list rather than a field and a null check per collaborator: migration, then legacy adoption, then loose import, in the order the bootstrap adds them. Ordering is a comment where the order is chosen. Phase 3 adds catalogue adoption and pack installs to the same list.

`Registry::mpdf_font_config()` exists because two places construct mPDF — `Helper_PDF::begin_pdf()` and the v3 `mPDF` shim in `deprecated.php` — and the shim had already drifted, missing the three keys 7.0 treats as requirements. Both now take the same array and merge their own per-document keys on top. `Test_Deprecated` gained the first test that constructs the shim at all.

**Catalogue integrity — four recorded defects, three fixed and one withdrawn**

These sit inside phases that already merged, so each is a fix on top of its phase rather than a change to what the phase specified. The full write-ups are §11 of the plan.

*Fixed.* `validate_entry()` checked a file's name and `remote_path` and stopped, leaving `sha256` and `size` — the two fields the whole install-time security model rests on — unvalidated. Never exploitable, since the runtime fails closed either way, but it moved a build mistake from a refused index at sync onto a customer's site as a `font_size_mismatch` mid-install. Both are required now, and `size` must be a real integer rather than merely numeric: the published artefact emits one, and accepting `"10"` is accepting a build that has started stringifying its numbers.

`Registry::installed_package()` passed an empty alias map, so `font-family: eeyekunicode` in a 6.x template resolved to nothing after the 7.0 key renames — no error, no fallback notice, the PDF simply drawn in the document font. The same silent shape as GF 3.0 dropping `RGForms`. An entry now publishes `aliases` as `alias => font key`, validated by the same table `language_to_font` goes through. The empty argument stays, and that is the fix rather than a shortcut: mPDF merges every package's aliases into one `fonttrans`, so only one list needs building and it has to be the one that sees every row — which is what enforces an alias deferring to a real font of the same name. The inversion sits beside the coverage maps rather than inside them, because a rename is something any source can do and Google has renamed a published family; gating it on coverage would have let a display family's rename pass validation and then be dropped at row-write time.

`sync_root()` stamped only `last_attempt` when a source's content had not changed, so `synced` tracked the store's last *publish* rather than this site's last successful *verification* — and staleness reads `synced`. Thirty quiet days would have put the whole fleet on the retry backoff and raised the catalogue health check against a store answering perfectly, training every admin to ignore the one signal that would show a freeze attack. A verified root that still names the index a site holds now counts as a sync and clears the stale error line. The stamp sits below the changed-check, not above it: stamping first would mark a source fresh whose index then failed to arrive or parse, which is the same defect pointed the other way.

*Withdrawn.* The fourth was to refuse an entry whose `remote_path` basenames collide — the residue of five families' `LICENSE.txt` collapsing to one file. Written, then taken back out: it fires on exactly the wrong entries. `remote_path` only builds a download URL, and the local name is the `files` map key, which is already unique — so two licences at `a/LICENSE.txt` and `b/LICENSE.txt` under distinct keys are lossless, and the rule refused them, while the entry the defect is actually about still validated, because the files the builder dropped left no trace to compare against. Since one bad entry refuses a whole source index, that is a third-party catalogue lost for a well-formed index. The invariant is only checkable where both sides of the collision still exist, which is the build pipeline.

Verified against the live catalogue rather than fixtures alone: all seventeen published packs pass the tightened validation unchanged.

**Phase 6 — the language settings**

`Rest_Font_Settings` serves the last entry in the client's `PENDING_ROUTES`, so that array, `mockOnly()` and `enablePendingRoutes()` are all deleted: `api/mock/` is the Jest seam now and nothing else.

`Registry::language_map()` is the screen's view of the two maps §4.4 already computes — one row per code in `$default ∪ $effective`, grouped the way `get_grouped_fonts()` groups the sidebar so a label and an order can only be wrong in one place. It files a row under its *default* font rather than its current one, so a code moved elsewhere stays with the pack that claimed it and an admin can see what they changed it from. "Changed" stays derived (`font !== default_font`) rather than becoming a second fact that can disagree with the first, and the overrides are not sent back at all, for the same reason.

Three deviations from what §4.4 and §4.8 specified, each recorded in the plan:

- **The sentinel is `*` on the wire in both directions, not `''`.** The panel's font select carries a `*` option and no empty one, so an `''` would have drawn as "no selection" and saved back as whatever option happened to be first. `Registry::LANGUAGE_NONE` is the one spelling now, shared with `effective_language_map()` and with the client through `constants.js`.
- **`document_script` stores the bare script name** (`LATIN`, `HAN`) — what the select offers. `get_document_script()` was looking up `Ucdn::LATIN`, which mPDF does not define, so the setting had never once taken effect. `Registry::script_constant()` owns the vocabulary now: it folds mPDF's `SCRIPT_` prefix either way, and the route refuses a name no constant answers to rather than storing a `BANANA` that looks like a setting and does nothing. `Registry::scripts()` ships the offerable list so the client's third copy of it could be deleted.
- **The four keys are deliberately not registered as settings fields.** §4.8 asked for that so they would be "sanitised/exported with the rest" — they would be neither. `settings_sanitize()` is tab-scoped and there is no settings export tool, while a section the sanitiser *can* reach has its unanswered fields blanked on every save of that tab, which for `font_language_overrides` means an unrelated screen wipes the override map. Naming the section something no tab is a prefix of avoids that only by coincidence of strings: the tab comes from the referer with no whitelist, and the tab list is filterable. `Rest_Font_Settings` is the only writer and `Registry` the only reader and defaulter. Per-PDF `pdf_language` **is** registered, in `form_settings_appearance`, where the framework genuinely carries it — rendering, sanitising and a REST enum.

`Model_Form_Settings::settings_sanitize()` no longer writes `[]` for an unanswered single-value select. It had since 6.x, for `template`, `pdf_size`, `orientation` and `font`. `multicheck` still gets an array, and so does a select declaring `multiple`, which core has none of and a third-party template may. `Registry::setting()` stays regardless — sites carry years of `[]` already on disk, and the reader is what makes those readable; it replaces the one-off `is_string()` guard `Coverage_Resolver` had grown for the same reason.

`prune_language_overrides()` lives on `Registry` beside `effective_language_map()`, its read-side twin, rather than in the controller that posts to it: the upgrade routine, a repair pass and WP-CLI all have to be able to write a valid override map without restating the rule. `default_language_map()` is memoised on the repository stamp, since three callers of one request were each re-sorting every installed row to rebuild it.

`Language_To_Font::labels()` is a table of ~190 language codes and ~59 `und-<script>` ones — the union of every code a published pack claims, every code the frozen 6.x map carries, and the Latin-script languages the default-language select needs but no pack has to. It is memoised per locale, because one of its callers is `get_registered_fields()` and that runs on **every REST request the site serves**, `GET /download/{entry}/{pdf}` included; the locale key is what keeps a notification email rendering in the recipient's language rather than the site's. An unnamed code degrades to the raw code, and a test locks the one code set knowable in advance — every language a 6.x upgrade can adopt off disk.

`Test_Language_Behaviour` pins §4.4's behaviour table against a real mPDF, asserting the fonts actually loaded rather than the map's own arithmetic: a Japanese document keeps its chosen font for Han and kana, the same document in English loses its Han to the installed pack, `und-Hans → *` is the way back, an explicit `lang` attribute reaches the map on any document, Vietnamese is not flipped out of the document font, and Cyrillic stays put in a font that carries it but is substituted from Arimo in one that does not.

**Deferred on purpose**

The `entry` pack linkage is not in the frozen map: the pack table is Phase 3a, and every adopted family lands as an imported row under its frozen 6.x key. That is the correct end state for an upgraded site regardless — the Noto swaps are only meant to reach fresh pack installs — so Phase 3a's catalogue adoption has nothing to undo.

The `Health_Issue` raised on a failed `dbDelta` goes in with the rest of the health framework; the `gfpdf_health_report` and `gfpdf_font_catalog_root` uninstall lines go in with the features that create them.

**Still to come**

The airgapped import path (§10), where a site with no catalogue verifies a package's detached signature instead. The first WP-CLI command, `wp gpdf fonts install <source>/<entry> [--from=<dir|zip>]`, which is the third offline route the docs page describes. Then Phase 7 — the distribution readme carrying the `fonts.gravitypdf.com` disclosure, the privacy-policy content, the upgrade notice and changelog, the docs pass, and the add-on compatibility sweep.

**Known, and deliberately carried**

**`Co-Authored-By` trailers are incomplete, and by more than this description previously claimed.** Two commits were named here as the only ones missing the trailer; both have since been fixed. An audit of the whole branch then found the real figure: **34 of the 89 commits carry no trailer**, of which 18 are `docs:` and `build:` and **16 are code** — 14 `feat:` and 2 `refactor:`, a contiguous block dated 9–10 September that is precisely the Phase 3 work (the catalogue table and sync, the downloader, the install queue, the REST routes, the install triggers, the health check and System Report, and multisite ownership). They are left alone pending a decision on each, since a trailer is a claim about who wrote something and is not worth guessing at. The force-push constraint that once justified deferring this no longer applies — the rebase onto #1727 already did one. The `package` column's identity design has an alternative worth weighing before the first publish, written up at `.claude/plans/2026-09-14-entry-identity-in-the-archive.md`. And the store side owes two gates before display-family packages can ship, written up for hand-off in the update-server repo.

Two obligations now sit with the pipeline rather than here. The `aliases` field has a consumer waiting but nothing publishing it yet, so no pack can rename a key until it does. And the shared accept/reject corpus the pipeline's schema gate runs against (`fonts/pipeline/expected/entry-schema.json` in the update-server repo) needs the new hash and size cases before the next publish — that gate calls this branch's `validate_entry()` out of a Composer package, so the two will disagree until it does.
</details>


src/Controller/Controller_Upgrade_Routines.php
```diff
@@ -101,6 +109,20 @@ protected function update_background_processing_values(): void {
...
+	 *
+	 * The notice and its route are gone, so the entry is dead weight in an autoloaded option.
+	 *
+	 * @since 7.0
+	 */
+	protected function clear_core_font_dismissal(): void {
+		/** @var \GFPDF\Model\Model_Actions $model */
+		$model = \GPDFAPI::get_mvc_class( 'Model_Actions' );
+
+		$model->undismiss_notices( [ 'install_core_fonts' ] );
```
jakejackson1: Don't think it could be dismissed.

PR #1725, opened 2026-09-08
Remove the version 3 compatibility layer for 7.0

jakejackson1:
## Summary

Gravity PDF has carried a version 3 compatibility layer since 4.0 shipped in 2016. This removes it for 7.0.

Everything here was deprecated a decade ago and, in most cases, deprecated silently — the v3 hooks fired with no notice attached at all, and a v3 template rendered a blank PDF rather than an error. That is why 6.17 went out first with detection: the System Status report, a Site Health test and a one-time admin notice each tell a site exactly which forms, templates and hooks are affected before anything stops working. This branch is the other half of that plan.

What goes: the global `src/deprecated.php` classes (`GFPDFEntryDetail`, `PDF_Common`, `PDFRender` and friends), the `?gf_pdf=1` download URLs, v3 templates and the Advanced Templating (Business Plus / Tier 2) support built on them, the `gfpdfe_*` filters and their v3-shaped `gfpdf_*` aliases, and roughly 35 stub methods that have been empty or one-line shims for four or more minor releases.

None of it fails silently on the PHP side. Every class that lost a method gains `__call()`/`__callStatic()` from a new `Helper_Trait_Removed_Methods`, so a stale caller gets a `_deprecated_function()` notice and a `null` rather than an "undefined method" fatal. It is a report, not a shim — the call still has to go — but a site finds out from its deprecation log instead of a white screen.

Two things deliberately **stay**, because PDF templates rely on them and always will: the `PDF_TEMPLATE_LOCATION` family of path constants and the `$gfpdfe_data` global. What changed is when they exist — they are now defined once a PDF render begins rather than on every request. Code that read one at `init` needs `GPDFAPI::get_data_class()` instead.

The one that needs saying plainly: **legacy download URLs do not 404.** `gf_pdf` was never a registered query var, so with the endpoint gone WordPress sees no route and serves the front page with a 200. A link in an already-sent notification email quietly stops producing a PDF and lands somewhere that looks fine, leaving the customer nothing to report. That has to lead the upgrade guide.

⛔ **Before this is deployed anywhere:** three sites we own (`gravity-pdf-website`, `site-gravitypdf-demo`, `cover-letter-website`) repoint their template directory through the v3 `gfpdfe_template_location*` filters in an mu-plugin. Each falls back to the default template directory the moment 7.0 reaches it. The one-line swap to `gfpdf_template_location` / `gfpdf_template_location_uri` is safe to make today on 6.x.

## Try it

1. Check out the branch and load any form with a PDF configured. Generating, viewing, downloading and emailing a PDF should be unchanged.
2. Visit **Forms → System Status**. On a clean site there is no Deprecated Features section at all. On a site still using something removed, an **Unsupported** section lists each affected template file and form ID, linking to that form's PDF settings.
3. Drop a file with no `Group:` header into `wp-content/uploads/PDF_EXTENDED_TEMPLATES/`, then reload the system report — it is reported as a legacy template, by file name, against the forms configured to use it.
4. Follow an old `?gf_pdf=1&fid=1&lid=1&template=zadani.php` URL. You get the front page with a 200, not a PDF and not a 404 — the behaviour the changelog and the upgrade guide have to describe.
5. **Tools → Site Health** carries the same detections, and the dashboard shows one dismissible error notice covering all of them.
6. With `WP_DEBUG` on, call something that is gone — `GPDFAPI::get_pdf_class( 'model' )->cleanup_pdf( [], [] )`, say. The request finishes and `debug.log` gets one "deprecated since version 7.0" line, where before it would have been fatal.

## Review pass

All ten comments from the first-pass review are resolved in the branch as it now stands. Four of them put something back that the removal had taken out: the `gfpdfe_*` hook detection (now reported as removed rather than deprecated), `Deprecation::apply_filters()` and `::mark_feature_detected()`, and `View_PDF::generate_pdf()` (delegating to `\GPDFAPI::create_pdf()`). `$legacy_ids` stays on `get_template_arguments()` rather than being dropped, so 7.0 breaks no signature, but it is marked deprecated and no longer read — `lead_ids` is hardcoded to the entry being rendered, which is the only value the argument could have produced once the multi-entry endpoint went. The CI gate is gone.

One comment went around twice. `Helper_Abstract_View::__call()` routes any unresolved name to a template of that name, so with `View_Settings::tabs()` removed and `Settings/tabs.php` still called that, a caller on the old method half-rendered the nav — two PHP warnings and an empty `<nav>` — rather than being told the method was gone. Keeping the filename meant declaring removed names for `__call()` to answer first; renaming the partial to `sub_menu.php` drops the collision instead, so that declaration and the guard reading it both come back out. The file now matches `sub_menu()`, the only method that renders it.

One comment produced no code change. It asks for the legacy-template refusal in `Helper_PDF::set_template()` to be shown only to site admins, and it already is — `Model_PDF::process_and_save_pdf()` is the sole production caller, it catches the exception into the log and returns a fixed `pdf_generation_failure`, and `Controller_PDF::pdf_error()` gates that on `gravityforms_view_settings`. A capability branch was written and taken back out: its only observable effect was to blank the template name in the log whenever no privileged user is present, which is cron, async notifications, REST and WP-CLI — exactly when support needs it. A comment on the throw now records that, so the next read does not have to re-derive it. Happy to add the gate anyway if you would rather have it.

The four removal notices were reworded after the review pass, for plain language and to say what a site actually loses rather than naming internals. They read, in the System Status report and the admin notice both:

> PDFs that use a legacy template can no longer be generated in Gravity PDF 7.0. Rebuild those templates in the current format.
> PDFs built with Advanced Templating can no longer be generated in Gravity PDF 7.0. Contact GravityPDF.com to discuss your upgrade options.
> Old PDF download links stopped working in Gravity PDF 7.0, so anyone who clicks one will not get their PDF. Replace those links with the `[gravitypdf]` shortcode or a PDF merge tag.
> Custom code on this site uses Gravity PDF hooks that were removed in version 7.0 and has stopped running. Update it to use the current hooks.

Each is followed by a "Learn how to upgrade" link on both surfaces, so the copy does not have to carry the how.

## Test plan

- PHPUnit, single site: 1608 tests, 5002 assertions, green
- PHPUnit, multisite: 1608 tests, 5099 assertions, green
- Playwright `core-isolated` (the deprecated-features surfaces): 4/4 green, but taken before the review pass restored the hook assertions and before the notice rewording — those four need re-running
- Playwright `core` (full project, 66 tests): 45 green. The 21 failures are all pre-existing — 20 in `pdf-settings.spec.ts` and one in `page-confirmation.spec.ts`, both reproduced identically with the branch stashed. Across that whole run the debug log records no new deprecation notice, so nothing in the plugin reaches its own safety net
- PHPCS over `src/` and `tests/`: clean
- The `Template_Constants` once-guard mutation-tested: removing `static::$defined` fails that test and only that test

<details>
<summary>Technical detail</summary>

### Shape of the branch

One commit against `development`, covering the eight phases of `.claude/plans/2026-08-17-7.0-deprecated-code-removal.md` plus the review pass. It overlaps open PR #1701, which decouples the localised script data from `PDF_TEMPLATE_LOCATION` — if that merges first, this rebases onto it cleanly.

### The path constants are now lazy

`GFPDF_Core::setup_constants()` used to define `PDF_SAVE_LOCATION`, `PDF_FONT_LOCATION`, `PDF_TEMPLATE_LOCATION` and `PDF_TEMPLATE_URL_LOCATION` on every request from `deprecated.php`. They now come from `Statics\Template_Constants::maybe_define()`, called from `Helper_PDF::__construct()` — so they exist for the whole of a render and are absent otherwise.

The once-guard matters and is not decoration: `$GLOBALS['gfpdfe_data']` has to stay identity-equal to the `Helper_Data` instance, or a template that mutates it on the first render loses those changes on the second. `define()` idempotency makes the constants useless as a probe for this, so the regression test asserts on object identity instead.

### The deprecation engine

`Statics\Deprecation` is a registry; `Deprecation_V3` is its one provider. The removal took the hook-warning half out along with the hooks it warned about; the review pass puts it back, so `apply_filters()`, the listener count, the development-site check and the log line are all still there, carried unused as the API the next round of deprecations will need. The detection half was never in question and now reports the removed `gfpdfe_*` hooks under Unsupported Features: a callback still attached to one has silently stopped running, which is precisely what a site needs told.

Every feature the v3 provider declares is now `GROUP_UNSUPPORTED`, so no provider declares `GROUP_DEPRECATED` and `get_groups()` omits it — the report section, the Site Health heading and the `gravity-pdf-deprecated` Info panel stop being registered rather than rendering empty. The group's label and description are deliberately kept in `View_System_Report::get_deprecated_groups()`: two callers index `['label']` unguarded, so deleting the copy would leave the first 8.0 provider that declares the group rendering an undefined index. That leaves five branches unreachable until Tier F, which is recorded as an explicit decision rather than left to drift.

### Filter chain order

`gfpdfe_template_location` and `gfpdfe_template_location_uri` ran **before** their v4 counterparts, so a site hooking both fed the v3 result into the v4 filter. `gfpdf_template_location` / `gfpdf_template_location_uri` now receive the plugin's own default directly — which is what the v3 pair used to be seeded with, so a site on the v4 filters alone sees no change.

`gfpdf_current_pdf_settings_object` **stays**. It is public API and still applied in `Model_PDF::generate_and_save_pdf()`; only core's own listener on it (`apply_backwards_compatibility_filters()`, via `Controller_PDF`) is removed. A side-effect worth having: that listener did its own `get_form()` through `gfpdf_current_form_object`, so every generated PDF was running `register_page_fields()`, the Gravity Perks conditional-date conversion and the Populate Anything hydration twice.

### CI gate

A grep gate in the `prebuild` job, failing the build if the removed classes came back, was added mid-branch and removed again on review. Recorded here because the reasoning may be worth revisiting when the 8.0 sweep lands: it belonged in `prebuild` rather than the `test-php` matrix, because a static fact about `src/` would otherwise be re-derived per PHP version and report as N identical red cells; and it inherited the repo's `run-tests` label gate, so an unlabelled PR merged without it and nothing ran on push to `development`. Making it always-on is a CI policy change and was left alone.

### The removed-method safety net

`Helper_Trait_Removed_Methods` supplies `__call()`/`__callStatic()` that run `_deprecated_function()` and return `null`. A `use` line on a class is the record that *this* class lost methods in 7.0, so the trait is opt-in rather than sitting on the MVC abstracts — blanketing those would turn an ordinary typo in any controller or model into a silent `null` blaming a removal that never happened.

`pdf.php` requires the trait by hand, next to the two files that block already loads: `GFPDF_Major_Compatibility_Checks` is declared before the autoloader is registered.

The views are the exception. `Helper_Abstract_View::__call()` is `final` and forwards any unknown method to a same-named template, so a trait method cannot override it — and it was already non-fatal, just silent. It reports through `_doing_it_wrong()` when the name resolves to neither a method nor a template, and still returns the `WP_Error`. The message is version-neutral: a mistyped template name reaches the same branch, and telling that author about a 7.0 removal would be a misdiagnosis. That is also why `Settings/tabs.php` is renamed rather than guarded — a view's template filenames are part of its public method namespace, so a future clash between a removed method and a partial is answered by renaming the partial.

Two consequences worth naming. Returning `null` is not a shim — a filter callback routed through `__call()` hands back `null` in place of the value it was given. And `is_callable()` now answers true for any name on these classes (`method_exists()` is unaffected), which changes one live guard: `Helper_Pdf_Queue::task()` gates on `is_callable()`, so a queue item left over from 6.x naming `Queue_Callbacks::cleanup_pdfs()` now runs through `__callStatic()` and the queue carries on, where before it was logged as critical and the queue cancelled.

This reverses one decision made partway through the branch, which deleted `Router::__call()`/`::__callStatic()` so an unknown Router method would fatal. The trait restores the notice under a 7.0 date and no changelog bullet announces the fatal.

### What the removals stranded

Cutting the v3 layer left `Controller_PDF`, `Model_Install` and `View_PDF` holding constructor arguments nothing reads any more. They were dropped and then put back. They are `protected` on non-`final` classes, so an add-on subclass may read them, and nothing in the org does — which means removal breaks a signature for no beneficiary. `Model_Install::$queue` and `Controller_Form_Settings::$form` are write-only independently of this branch, so pruning only the ones a v3 sweep happened to orphan would have been arbitrary. Auditing all of them is a container audit with its own breaking change, not part of a deprecation sweep. All three constructors now match `development` byte for byte and 7.0 breaks none of them.

`Helper_Templates::get_template_arguments()` keeps `$legacy_ids` for the same reason — the signature does not break — but the argument is optional, marked deprecated and no longer read. The legacy endpoint was the only caller that could pass more than one entry, so `lead_ids` is derived from `$entry` and the template argument itself is unchanged. A caller still passing the argument is ignored silently rather than warned; say the word if you would rather it raised `_deprecated_argument()`.

The `?template=` override gate had been trimmed to byte-identical copies in `Helper_PDF::set_template()` and `Helper_Templates::get_config_class()` — the same capability check deciding which template renders and which config class loads alongside it. It is now `Helper_Templates::get_requested_template_id()`, so the two cannot resolve differently.

### Selects pointing at something that is gone

Dropping v3 templates from the selector left any PDF still set to one rendering a `<select>` with no matching option, and a browser resolves that by selecting the first one — so opening the PDF's settings and saving reassigned it, silently. `build_options_for_select()` now prepends an option for a stored value it cannot find, labelled *not currently available*. That method rather than `select_callback()`, because the template manager rebuilds the same select over AJAX through `Model_Templates::ajax_process_build_template_options_html()`, which calls it directly.

The mechanism is not template-specific on purpose — a deleted font, and a template requiring a newer Gravity PDF than the site runs, have always rebound the same silent way. The option is left enabled, so saving without touching it preserves the value; the classic settings save sanitises rather than checking membership, so it round-trips.

One case could not simply be shown. `form_settings.template` takes its `std` from the global Default Template, and `get_form_value()` falls back to `std` for a PDF that has none yet, so a site whose default pointed at a v3 template would have handed that dead value to every PDF created after the upgrade. New PDFs start on `zadani` instead; the Global Settings screen still shows the dead default, which is where it gets corrected.

### Still outside this branch

- Publish the consolidated 7.0 upgrade guide on docs.gravitypdf.com, link it from the update notice, and add the URL to the top of the `### 7.0.0` changelog block
- The four doc pages in `gravity-pdf-documentation` that describe removed behaviour
- Bump `Version:` and `PDF_EXTENDED_VERSION` in `pdf.php` from `6.16.0` when 7.0 is cut
- The three mu-plugin filter swaps above; confirming the Previewer's `gfpdf_legacy_save_path` callback; migrating the demo site's 21 `advanced_template` PDFs

</details>


.github/workflows/tests.yml
```diff
@@ -62,6 +62,29 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v6
 
+      # Locks in the 7.0 v3 removal — the compatibility layer is gone and must not come back through a merge.
```
jakejackson1: Unnecessary CI check

src/Helper/Helper_PDF.php
```diff
@@ -370,32 +361,20 @@ public function save_pdf( $raw_pdf_string ) {
...
 
-		/* Check if there are version requirements */
 		$template_info = $this->templates->get_template_info_by_path( $this->template_path );
 
+		/* Refused rather than left to load_html(), which would include a v3 template whose Gravity Forms scaffolding
+		   is gone: its boilerplate returns early and mPDF writes a blank PDF the caller reads as a success */
 		if ( $this->templates->is_legacy_template( $template_info ) ) {
-			Deprecation_V3::restore_v3_form_class();
+			/* translators: 1: PDF template name wrapped in <em> tags, 2: URL of the legacy template upgrade guide */
+			throw new Exception( sprintf( esc_html__( 'The PDF Template %1$s is a legacy (v3) template, which Gravity PDF no longer renders. Rebuild it as a v4 template: %2$s', 'gravity-pdf' ), '<em>' . esc_html( $template ) . '</em>', esc_url( 'https://docs.gravitypdf.com/upgrade/legacy-templates/' ) ) );
```
jakejackson1, line 377: Only show to site admins.

src/Helper/Helper_Templates.php
```diff
@@ -749,19 +741,18 @@ public function get_template_image( $template, $type = 'url', $include_core = tr
...
+	 * @param array  $form_data The form data array, formatted form the $entry array
+	 * @param array  $settings  PDF Settings The current PDF settings
+	 * @param object $config    The current PDF template configuration class
 	 *
 	 * @return array
 	 *
 	 * @since 4.1
 	 */
-	public function get_template_arguments( $form, $fields, $entry, $form_data, $settings, $config, $legacy_ids ) {
+	public function get_template_arguments( $form, $fields, $entry, $form_data, $settings, $config ) {
```
jakejackson1: Make $legacy_ids optional

src/Helper/Helper_Trait_Removed_Methods.php
```diff
@@ -0,0 +1,51 @@
...
+	/**
+	 * Report a call to an instance method that no longer exists
+	 *
+	 * @param string $name      The method being called
+	 * @param array  $arguments The arguments it was called with
+	 *
+	 * @since 7.0
+	 */
+	public function __call( $name, $arguments ) {
+		_deprecated_function( esc_html( static::class . '::' . $name ), '7.0' );
```
jakejackson1: What if we remove more things in a follow-up version?

src/Statics/Deprecation.php
```diff
@@ -290,209 +295,4 @@ public static function group_signals( array $signals ): array {
...
-	 * moved on stay quiet.
-	 *
-	 * @param string $hook_name   The deprecated filter to fire
-	 * @param array  $args        The arguments to pass to the filter, the first of which is returned
-	 * @param string $replacement Override the mapped replacement, for dynamic hooks
-	 *
-	 * @return mixed
-	 * @since 6.17.0
-	 */
-	public static function apply_filters( string $hook_name, array $args, string $replacement = '' ) {
```
jakejackson1, line 308: Don't remove. It's useful for future deprecations.

src/Statics/Deprecation.php
```diff
@@ -174,22 +178,23 @@ protected static function store_detected_features( array $signals ): void {
...
+	 * @param string[] $keys The feature IDs to weigh
+	 *
+	 * @since 7.0
 	 */
-	public static function mark_feature_detected( string $key ): void {
-		$detected = static::get_detected_features();
-
-		if ( in_array( $key, $detected, true ) ) {
-			return;
+	public static function has_unsupported_feature( array $keys ): bool {
```
jakejackson1, line 209: Keep `mark_feature_detected`

src/Statics/Deprecation_V3.php
```diff
@@ -24,6 +25,8 @@
  */
 class Deprecation_V3 implements Helper_Interface_Deprecated_Features {
```
jakejackson1, line 26: Don't remove all this. Keep the detections, but instead of deprecated show it as removed.

src/Statics/Template_Constants.php
```diff
@@ -0,0 +1,80 @@
...
+		];
+
+		foreach ( $constants as $name => $value ) {
+			/* Guarded individually so a third party that got there first doesn't trigger a redefinition notice */
+			if ( ! defined( $name ) ) {
+				define( $name, $value );
+			}
+		}
+
+		$data->template_site_location = $template_url;
```
jakejackson1: Remove these two $data items (unless there's a good reason to keep)

src/View/View_PDF.php
```diff
@@ -105,162 +79,25 @@ class View_PDF extends Helper_Abstract_View {
...
-	 * Legacy view/download PDF generator
-	 *
-	 * @param array $entry    The Gravity Forms Entry to process
-	 * @param array $settings The Gravity Form PDF Settings
-	 *
-	 * @return void
-	 * @since      4.0
-	 * @deprecated 6.12.0 Use \GPDFAPI::create_pdf() to generate PDFs
-	 */
-	public function generate_pdf( $entry, $settings ) {
```
jakejackson1, line 155: Re-add, but route to \GPDFAPI::create_pdf().

src/View/View_Settings.php
```diff
@@ -158,7 +141,7 @@ public function sub_menu() {
 			'data'     => $this->data,
 		];
 
-		$this->load( 'tabs', $vars );
```
jakejackson1, line 161: Don't change the name

jakejackson1 (commented):
Bunch of changes needed

src/Helper/Helper_Templates.php
```diff
@@ -755,13 +747,14 @@ public function get_template_image( $template, $type = 'url', $include_core = tr
 	 * @param array  $form_data  The form data array, formatted form the $entry array
 	 * @param array  $settings   PDF Settings The current PDF settings
 	 * @param object $config     The current PDF template configuration class
-	 * @param array  $legacy_ids An array of multiple entry IDs for legacy templates only
+	 * @param array  $legacy_ids The entry IDs to advertise as `lead_ids`, which defaults to the entry being rendered
```
jakejackson1: Mark deprecated with no usage

src/Helper/Helper_Templates.php
```diff
@@ -779,8 +772,8 @@ public function get_template_arguments( $form, $fields, $entry, $form_data, $set
 			[
 
 				'form_id'   => $form['id'], /* backwards compat */
-				'lead_ids'  => $legacy_ids, /* backwards compat */
-				'lead_id'   => Deprecation::apply_filters( 'gfpdfe_lead_id', [ $entry['id'], $form, $entry, $gfpdf ] ), /* backwards compat */
+				'lead_ids'  => $legacy_ids ?: [ $entry['id'] ], /* backwards compat */
```
jakejackson1: Hardcode [ $entry['id'] ]

src/Statics/Deprecation_V3.php
```diff
@@ -132,52 +87,56 @@ class Deprecation_V3 implements Helper_Interface_Deprecated_Features {
...
-			'legacy_templates'              => [
+			'legacy_templates'        => [
 				'label'      => __( 'Legacy Templates', 'gravity-pdf' ),
-				'group'      => Deprecation::GROUP_DEPRECATED,
+				'group'      => Deprecation::GROUP_UNSUPPORTED,
 				'removed_in' => static::REMOVED_IN,
 				'url'        => 'https://docs.gravitypdf.com/upgrade/legacy-templates/',
 				'detect'     => [ static::class, 'get_legacy_templates' ],
+				/* translators: %s: The Gravity PDF version the feature was removed in */
+				'notice'     => __( 'Legacy (v3) PDF templates stopped rendering in Gravity PDF %s. Rebuild them as v4 templates.', 'gravity-pdf' ),
```
jakejackson1: Fix message

PR #1732, opened 2026-09-16
fix: note a PDF left off a notification, and log generation errors without the whole object (6.17)

jakejackson1:
When a PDF fails to generate for a notification, Gravity Forms still sends the email, just without the attachment, and nothing outside the debug log says so. The entry now gets an error note naming the PDF and the notification, so the missing attachment is visible without logging turned on.

The PDF generation error logs also no longer dump the whole PDF generator, entry or PDF settings. On a customer's site that came to 115 KB per error. They now record the form, entry and PDF IDs, the template, the path and the exception's message, file and line.

Both came out of the same customer investigation as #1731, where two PDFs failed with mPDF's `Temporary files directory ... is not writable` and the admin emails went out without them.

## Try it

Pick an entry whose form has a PDF attached to one of its notifications, then force the PDF to fail:

```bash
wp eval '
add_filter( "gfpdf_pdf_config", function ( $pdf ) { $pdf["template"] = "doesntexist"; return $pdf; } );
$entry = GFAPI::get_entries( 0, [], null, [ "page_size" => 1 ] )[0];
$form  = GFAPI::get_form( $entry["form_id"] );
foreach ( $form["notifications"] as $notification ) {
	GPDFAPI::get_mvc_class( "Model_PDF" )->notifications( $notification, $form, $entry );
}
print_r( wp_list_pluck( GFAPI::get_notes( [ "entry_id" => $entry["id"], "note_type" => "gravity-pdf" ] ), "value" ) );
'
```

Each notification the PDF is attached to adds a note like `The PDF "My PDF" could not be generated and was not attached to the "Admin Notification" notification.` On 6.17.0 the list is empty.

## Test plan

- [ ] A notification whose PDF fails to generate still sends, and the entry gets the note above
- [ ] A notification whose PDF generates normally adds no note
- [ ] With logging on, a failed PDF logs IDs and the exception, not the full object, for both notifications and view/download
- [ ] `Test_PDF` passes

<details>
<summary>More info</summary>

**The note.** `Model_PDF::add_attachment_failure_note()` calls `GFFormsModel::add_note()` with type `gravity-pdf`, subtype `error`, user `Gravity PDF`. It lives in `notifications()` because that is the only path where a failed PDF is silently dropped; the API, `maybe_save_pdf()` and the background queue don't send email. The PDF and notification names are escaped, since Gravity Forms prints notes through `wp_kses_post()`.

**Log context.** `Model_PDF::process_and_save_pdf()` and `View_PDF::generate_pdf()` log `form_id`, `entry_id`, `pdf_id`, `template` (plus `path` in the model), `exception`, `file`, `line`. `View_PDF::process_field()` logs `form_id`, `entry_id`, `field_id`, `field_type`, `exception`, `file`, `line`. Secrets in the old dumps were already redacted by `Redact_Processor`; the change is about size and entry data.

**Tests.** A new `test_notifications_notes_a_pdf_that_failed_to_attach` breaks the template through `gfpdf_pdf_config` and asserts the empty attachment list, one entry note, and a log record carrying `entry_id` without the object. It fails against 6.17.0 and passes here. `Test_PDF`, `Test_View_PDF` and `Test_Slow_PDF` pass locally (95 tests, 1 skipped). PHPCS is clean.

</details>


src/Model/Model_PDF.php
```diff
@@ -1118,6 +1120,26 @@ public function notifications( $notifications, $form, $entry ) {
...
+	/**
+	 * Record on the entry that a PDF was left off a notification, which is otherwise sent without it
+	 *
+	 * @param array $entry        The Gravity Forms entry
+	 * @param array $settings     The Gravity PDF settings
+	 * @param array $notification The Gravity Forms notification
+	 *
+	 * @since 6.17.1
+	 */
+	protected function add_attachment_failure_note( array $entry, array $settings, array $notification ): void {
```
jakejackson1: Remove types

